PR #6, opened 2022-12-27
T5: Remove npm test

### LINKED ISSUE #5: Remove "npm test" from front end workflows

NasihNazeem:
I decided to remove npm test because we won't necessarily touch the App.test.js file so there's no point in keeping that command


Dennull:
Will address this in PR #4

PR #18, opened 2023-01-16
T15: Timeline post view

### LINKED ISSUE #15: Timeline posts
This is how users see posts on their timeline. It includes the post and/or post media, the like button, repost button, etc. MVP requirements:

- When the timeline loads, the component should make a GET request to the backend. The component should map the array of posts that comes back from the request (if the request was successful)
- The posts should be clickable and have buttons for liking, reposting, commenting, and sharing
   - These buttons can do nothing for now
- The post should show normal text or an image or both

Dennull:
Closes #15. Created `<PostList>`, which is the component that will render the posts that user sees on their timeline


NasihNazeem:
Looks good, I think if there's a way to remove the highlight or ripple effect for button clicks that would be better. Also need to create a function that for liking a tweet, retweeting. There should also be post requests to the database whenever you like, retweet, or undo the action

src/components/Post.tsx
```diff
@@ -11,7 +11,14 @@ import AddCommentOutlinedIcon from "@mui/icons-material/AddCommentOutlined";
 import RepeatOutlinedIcon from "@mui/icons-material/RepeatOutlined";
 import ShareOutlinedIcon from "@mui/icons-material/ShareOutlined";
 
-const Post = () => {
+type PostProps = {
+  displayName: string;
+  textContent: string;
+  timestamp: string;
```
NasihNazeem: This would be a type "Date"

src/App.js
```diff
@@ -1,13 +1,15 @@
-import "./App.css";
-import Login from "./pages/Login";
+import Welcome from "./pages/Welcome";
+import Timeline from "./pages/Timeline.tsx";
```
Dennull: Think we shouldn't need the ".tsx" at the end here

src/pages/Timeline.tsx
```diff
@@ -0,0 +1,26 @@
...
+const Timeline = () => {
+  return (
+    <Stack direction="row" sx={styles.root}>
+      <Box>
+        <ComposePost />
+      </Box>
+      <Box sx={styles.postListContainer}>
+        <PostList />
+      </Box>
+    </Stack>
```
Dennull: I don't think this is the layout we want. If we do it like this then the `CompostPost` component will show up to the left of the `PostList` component (on the same row). We want `ComposePost` to be above the `PostList`. I think it should look something like this:

```suggestion
    <Stack direction="row" sx={styles.root}>
      <Box sx={styles.postListContainer}>
        <ComposePost />
        <PostList>
      </Box>
    </Stack>
```

Dennull (commented):
@NasihNazeem think we need to make some small changes before merging. I pulled the branch and ran it though and looks like there aren't any conflicts!

NasihNazeem (approved):

PR #22, opened 2023-02-02
T21: Theme

### LINKED ISSUE #21: Theming cleanup
Theming will take care of the theme for all the components we are implementing that require an on-brand look. This includes things like buttons, modals, text font, etc.

NasihNazeem:
I am just going to merge this into release 1 so you guys get access to the current developed themes and then we can all add on to it depending on what we work on, that way the themes are not dependent on separate components being developed


Dennull:
Looks good!

Dennull (commented):
Looks good!

PR #25, opened 2023-03-22
T24: Navbar post button modal

### LINKED ISSUE #24: Navbar post button modal
When the user clicks the "post" button on the navbar, show a modal that allows them to post. The modal should appear from anywhere on the app.

NasihNazeem:
This closes #24. Additions in this PR:
- `PostButton` opens up a modal (MUI Dialog)
- Modal uses the `ComposePost` component as a child
- ModalProps includes children components, openModal and setOpenModal
- Theming changes to create a grey button in `Theme.tsx`

### DIFF
--- a/src/components/Misc/PostButtonModal.tsx
+++ b/src/components/Misc/PostButtonModal.tsx
@@ -1,29 +1,35 @@
-import React, { Dispatch, ReactElement, SetStateAction } from "react";
 import { Dialog, DialogTitle, DialogContent, Button } from "@mui/material/";
 
 const styles = {
-  dialogPosition: {
+  dialog: {
     maxHeight: "90%",
+  },
+  greyButton: {
+    color: "#D8D8D8",
   },
 };
 
-type ModalProps = {
-  children: ReactElement<any, any>;
+type PostButtonModalProps = {
+  children: JSX.Element;
   openModal: boolean;
-  setOpenModal: Dispatch<SetStateAction<boolean>>;
+  onClose: () => void;
 };
 
-export default function PostButtonModal(props: ModalProps) {
-  const { children, openModal, setOpenModal } = props;
+export default function PostButtonModal({
+  children,
+  openModal,
+  onClose,
+}: PostButtonModalProps) {
   return (
     <Dialog
       open={openModal}
       fullWidth
       scroll="paper"
-      sx={styles.dialogPosition}
+      sx={styles.dialog}
+      onClose={onClose}
     >
       <DialogTitle>
-        <Button color="greyButton" onClick={() => setOpenModal(!openModal)}>
+        <Button sx={styles.greyButton} onClick={onClose}>
           X
         </Button>
       </DialogTitle>
--- a/src/components/Misc/PostButtonModal.tsx
+++ b/src/components/Misc/PostButtonModal.tsx
@@ -1,11 +1,18 @@
-import { Dialog, DialogTitle, DialogContent, Button } from "@mui/material/";
+import { Dialog, DialogTitle, DialogContent, IconButton } from "@mui/material/";
+import CloseIcon from "@mui/icons-material/Close";
 
 const styles = {
   dialog: {
     maxHeight: "90%",
   },
-  greyButton: {
-    color: "#D8D8D8",
+  dialogContent: {
+    padding: 0,
+  },
+  dialogTitle: {
+    paddingBottom: 0,
+    paddingTop: 0,
+    paddingRight: 0.5,
+    paddingLeft: 0.5,
   },
 };
 
@@ -28,12 +35,12 @@ export default function PostButtonModal({
       sx={styles.dialog}
       onClose={onClose}
     >
-      <DialogTitle>
-        <Button sx={styles.greyButton} onClick={onClose}>
-          X
-        </Button>
+      <DialogTitle sx={styles.dialogTitle}>
+        <IconButton disableRipple onClick={onClose}>
+          <CloseIcon />
+        </IconButton>
       </DialogTitle>
-      <DialogContent>{children}</DialogContent>
+      <DialogContent sx={styles.dialogContent}>{children}</DialogContent>
     </Dialog>
   );
 }
--- a/src/components/Misc/PostButtonModal.tsx
+++ b/src/components/Misc/PostButtonModal.tsx
@@ -1,5 +1,23 @@
-import React from "react";
+import React, { Dispatch, SetStateAction } from "react";
+import { Dialog, DialogTitle, DialogContent, Button } from "@mui/material/";
 
-export default function PostButtonModal() {
-  return <div>PostButtonModal</div>;
+type ModalProps = {
+  openModal: boolean;
+  setOpenModal: Dispatch<SetStateAction<boolean>>;
+};
+
+export default function PostButtonModal(props: ModalProps) {
+  const { openModal, setOpenModal } = props;
+  return (
+    <Dialog open={openModal}>
+      <DialogTitle>
+        <Button color="greyButton" onClick={() => setOpenModal(!openModal)}>
+          X
+        </Button>
+      </DialogTitle>
+      <DialogContent>
+        <div>content goes here.</div>
+      </DialogContent>
+    </Dialog>
+  );
 }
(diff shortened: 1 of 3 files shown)


src/components/Misc/PostButtonModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const styles = {
+  dialogPosition: {
+    position: "absolute",
+    bottom: "500px",
+  },
+};
+
+type ModalProps = {
+  children: ReactElement<any, any>;
```
Dennull: Can we use a different type here? I think `ReactNode` or `PropsWithChildren` looks better. 

Also nitpick but I think `ModalProps` is a but vague, since we could have multiple different modals being used. The convention I use is the component name with "Props" at the end (eg. `PostButtonModalProps`), but we can also come up with our own

src/components/Misc/PostButtonModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+  openModal: boolean;
+  setOpenModal: Dispatch<SetStateAction<boolean>>;
```
Dennull: This type looks really gross. I think it's better to pass this as a callback function that gets invoked. We can then call it `onClose` or something similar and pass that to the modal to set the close state. The type for that would be a function signature like this:
```suggestion
  onClose: () => void;
```

In the parent component we'd pass in the prop like so:

`<PostButtonModal onClose={() => setOpenModal (false)}>`

src/components/Misc/PostButtonModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export default function PostButtonModal(props: ModalProps) {
```
Dennull: Not sure how everyone feels about this, but we could also destructure our props directly in the brackets:
```suggestion
export default function PostButtonModal({children, openModal, setOpenModal}: ModalProps) {
```

Comes down to preference but I think as long as we pick something and use it across the app

src/components/Misc/PostButtonModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const { children, openModal, setOpenModal } = props;
+  return (
+    <Dialog
+      open={openModal}
+      fullWidth
+      scroll="paper"
+      sx={styles.dialogPosition}
+      PaperProps={{
```
Dennull: Do we need this? What does it do?

src/components/Misc/PostButtonModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+        style: {
+          minHeight: "20%",
+          maxHeight: "90%",
+        },
+      }}
+    >
+      <DialogTitle>
+        <Button color="greyButton" onClick={() => setOpenModal(!openModal)}>
```
Dennull: Instead of using a button with an onClick handler, we should use the `onClose` prop on the dialog. They have some examples on MUI's website: https://mui.com/material-ui/react-dialog/#customization 

It not only helps with styling (the padding on the title is huge right now), but also makes it so that when the user clicks away from the dialog it closes too

Along with my `onClose` suggestion above, it would look something like this instead:

```suggestion
        <Dialog onClose={onClose}>
```

src/styles/Theme.tsx
```diff
@@ -18,6 +35,11 @@ const theme = createTheme({
     success: {
       main: "#22AA6F",
     },
+    greyButton: palette.augmentColor({
```
Dennull: I wanna see if there's a neater way to do this. I think we should try to reduce the amount of interfaces we have in this file. It also might not be worth making a new class in theme just for 1 grey button

src/components/NavBar/NavBar.tsx
```diff
@@ -70,12 +75,20 @@ const NavBar = () => {
...
+            type="button"
+            onClick={() => setOpenModal(!openModal)}
+          >
             Post
           </Button>
           <UserAvatar />
         </Stack>
       </Toolbar>
+      <PostButtonModal openModal={openModal} setOpenModal={setOpenModal}>
+        <ComposePost />
```
Dennull, line 94: I like that you're trying to reuse our components, but dialogs have a `DialogActions` component that we can put our post button, emoji upload button, etc inside. I think the styling might look better if we try to use the component that they have

Dennull (changes requested):

src/components/Misc/PostButtonModal.tsx
NasihNazeem: `PropsWithChildren` would fit better than `ReactNode` or `ReactElement`, but I decided to go with `JSX.Element` instead because we already know what `children` is. If the `children` prop was unknown or dependent on other components, I think the other two would be better fitting but in this case, we can be more specific since `ComposePost` is the only child prop we are expecting in a `PostButtonModal` component

src/components/Misc/PostButtonModal.tsx
NasihNazeem: I have applied this change.

src/components/Misc/PostButtonModal.tsx
NasihNazeem: I have applied this change. It is better to destructure. Cleaner to read through and saves a line of code.

src/components/Misc/PostButtonModal.tsx
NasihNazeem: **fullWidth**: Uses 100% of the parent's width

**scroll**: Has two options, ``paper`` and ``body``. 

> ``paper`` allows you to scroll through the ``DialogContent`` whilst fixing the location of the ``DialogTitle``
> ``body`` allows you to scroll through the ``DialogContent`` but this time you are scrolling through the entire body meaning the ``DialogTitle`` can be scrolled past.

**sx** and **PaperProps**: PaperProps is not necessary so I've removed it and decided to make a design choice for the positioning of the component. This component will now be placed in the middle of the page rather than closer to the top of the page.

src/components/Misc/PostButtonModal.tsx
NasihNazeem: I have applied this change due to changing the `setOpenModal` prop. It would be a better fit for the reasons you mentioned as well. 

One pushback is I am keeping the onClick handler because onClose on dialog does not affect the close button functionality. I am now calling onClick as `onClick = {onClose}` within the Button attributes.

src/styles/Theme.tsx
NasihNazeem: For us to include custom colours in Typescript we will need to customize these interfaces for it to recognize any custom colours that we create. 

But yeah honestly it's just one grey button.. If we want to use this colour again, or want to revisit augmentation then I think we can start looking into it. 

I resulted in removing the augmentation and passing in the `sx` prop on Button with a custom `styles` that provided that color.

src/components/NavBar/NavBar.tsx
NasihNazeem: It seems redundant to be remaking `ComposePost`, along with the post request and whatnot

src/components/Misc/PostButtonModal.tsx
Dennull: Nice! You're right, we still do need the `onClick` handler. I think we should also use MUI's `IconButton` component rather than their `Button` though

src/components/NavBar/NavBar.tsx
Dennull: On second thought, I agree. I think the padding still looks kind of bad though. Can we pair program and go through the styling together?

src/components/Misc/PostButtonModal.tsx
Dennull: As discussed offline, I don't really feel to strongly about which we use as long as we can avoid the use of the `any` type. We can stick with JSX.Element for now until we find a better reason to change it!

src/components/Misc/PostButtonModal.tsx
Dennull: Talked about this offline, we'll create a new tech debt ticket to handle some of the styling details

src/styles/Theme.tsx
Dennull: We went through this together and ended up getting rid of this color since it's only used in one spot. You're right though, if we end up adding new colors to our palette we'll have to use these interfaces

Dennull (approved):
Looks good, feel free to merge whenever!

PR #26, opened 2023-04-03
T8: User login and sign up

### LINKED ISSUE #8: User login and sign up
We'll be using Auth0 for authentication, so that takes care of the login and sign up forms. The bulk of this ticket is configuration and securing our front end routes. Requirements:

- Allow users to log in and out using Auth0
- When users are logged out, make sure the Auth0 login form is displayed if they navigate to any pages that require login

Additionally, after sign up, we need to display forms to collect additional data to store in our database. The forms should ask for:

- Username, which is unique and required
- Display name, which isn't required. We'll default to the user's username as their display name if one isn't entered
- Birthday, which isn't required

Users should not be able to click the submit button without fulfilling all the required labels. The user will continue to be shown the form when they log in until they fill in the required fields

Dennull:
This closes #8. This PR covers basic authentication configuration we need on the front end:

- Created an `AuthProvider` component which handles Auth0 configuration
- Created a `ProtectedRoute` component which will hide routes unless the user if logged in
- Set up `UserContext.tsx`, which exports a global store where the user's ID, display name, and username are stored
- `App.tsx` now conditionally renders the app if users are authenticated, and shows the welcome page if they aren't
- If a user doesn't have a username in our database (ie. when they first sign up), show the `Register` page to prompt them to fill out their username, display name, and birth date
  - Birth date and display name are optional. If the user doesn't include a birth date, it will be null. If the user doesn't include a display name, their username will be their display name


src/App.js
```diff
@@ -1,16 +1,68 @@
...
+        });
+        setUser(response.data);
+      } catch (error) {
+        console.log(error);
+      }
+    };
+    getUser();
+  }, [getAccessTokenSilently, setUser]);
+
+  if (isLoading || (isAuthenticated && user.isLoading)) {
```
NasihNazeem, line 39: We could maybe take these three if statements and create an if-else statement to make it easier to look at, but there would not be an else option, so I don't really see a point in that and we can just leave it as it is.

src/App.js
```diff
@@ -1,16 +1,68 @@
...
-        <Route path="/home" element={<Timeline />} />
+        <Route path="/" element={<ProtectedRoute component={Timeline} />} />
+        <Route
+          path="/messages"
+          element={<ProtectedRoute component={Messages} />}
+        />
+        <Route
+          path="/profile"
+          element={<ProtectedRoute component={Profile} />}
+        ></Route>
```
NasihNazeem: to keep consistent with the rest, I would just change this from `><Route>` to `/>`

src/components/Auth/AuthProvider.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const clientId = process.env.REACT_APP_AUTH0_CLIENT_ID;
+  const domain = process.env.REACT_APP_AUTH0_DOMAIN;
+  const redirectUri = window.location.origin;
+
+  const onRedirectCallback = (appState: any) => {
+    navigate(appState?.returnTo || window.location.pathname);
+  };
+
+  if (!(domain && clientId && redirectUri)) {
+    console.log("Authorization Provider Error!");
```
NasihNazeem: I think you mentioned this, but let's throw an error instead here

`throw new Error("Authorization Provider Error!");`

src/pages/Register.tsx
```diff
@@ -0,0 +1,125 @@
...
+      );
+      setUser(newUserInfo.data);
+    } catch (error) {
+      console.log(error);
+    }
+  };
+
+  return (
+    <form onSubmit={submitUserInfo}>
+      <Box sx={styles.container}>
```
NasihNazeem, line 65: We should look to center the form on this page, and potentially add things to it so it looks a bit more full.

Maybe like a low-opacity logo in the background.

src/pages/Register.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <Typography variant="h1" sx={styles.title}>
+          Let's get to know a little more about you
+        </Typography>
+        <TextField
```
NasihNazeem, line 69: This should be a unique identifier but I was able to post identical usernames onto this field. 

Ideally, there should be some check that queries the db to see if that username already exists, and if so, throw an error and do not redirect yet.

I can do this part later as part of another ticket also, looks like fun lol

NasihNazeem (changes requested):
I wrote some comments, but overall everything looked pretty good. 

I tested the functionality, and it worked as described. There were some QoL suggestions made, but the main functionality is there.

src/App.js
Dennull: If I remember correctly you can't use if-else for returning react components. It has to be an if

src/components/Auth/AuthProvider.tsx
Dennull: Yeah, good idea!

src/pages/Register.tsx
Dennull: Yeah I know, I was focused on getting the functionality to work first. I didn't put much validation. I'm open to making a new ticket to tackle that

src/pages/Register.tsx
Dennull: As discussed offline, we'll make another ticket to address this. An idea that came up was to use a modal to show the forms in

NasihNazeem (approved):

PR #37, opened 2023-05-10
T35: Redux

### LINKED ISSUE #35: Introducing Redux
We need to introduce Redux to this project to help manage the state of various objects in our program. 

This issue will mainly focus on the additions of:

- Redux Provider
- Redux Store
- Redux Reducer 
- Redux Dispatcher

We'll look to use Redux patterns to replace the UseContext hooks we implemented for User Authentication and Post Handling.

NasihNazeem:
This PR includes the introduction of Redux to the project. We have replaced all usage of UseContext for authentication and posts using Redux.


src/state/slices/userSlice.ts
```diff
@@ -0,0 +1,30 @@
+import { PayloadAction, createSlice } from "@reduxjs/toolkit";
+
+type UserState = {
+  value: User;
```
Dennull: Do we need this nested object behavior? Can we just ditch `value` and set the initial state to be a `user` object right away?

Same suggestion applies to `postSlice.ts`

src/App.tsx
```diff
@@ -10,13 +10,18 @@ import Register from "./pages/Register";
...
+import { useDispatch, useSelector } from "react-redux";
+import { setUser } from "./state/slices/userSlice";
 import { PostContextProvider } from "./context/PostContext";
 
 function App() {
   const { isLoading, isAuthenticated, getAccessTokenSilently } = useAuth0();
-  const { user, setUser } = useUserContext();
+  // const { user, setUser } = useUserContext();
+
+  const user = useSelector((state: RootState) => state.user.value);
```
Dennull: If we get rid of the nested `value` property like I suggest in `userSlice.ts` we can do something like this instead:

```suggestion
  const { isLoading, userId }= useSelector((state: RootState) => state.user);
```

I'd also like to create a custom hook that warps `useSelector` so that we don't have to keep passing the `state: RootState` type

src/components/Posts/PostItem.tsx
```diff
@@ -31,14 +33,14 @@ type PostProps = {
   post: Post;
 };
 
-const likePost = async (userId: number, postId: number) => {
+const likePost = async (userId: number | undefined, postId: number) => {
```
Dennull: It was working fine before since I hard coded this to be 1, but we shouldn't allow an undefined type here since the server will throw an error if undefined actually got passed

We either need to do an undefined check or make it so that the `userId` property on the `user` object can't be undefined. I'm inclined to do the second option because really a `user` object shouldn't be able to exist without a `userId`

src/state/store.ts
```diff
@@ -0,0 +1,14 @@
...
+export const store = configureStore({
+  reducer: {
+    // slices go here
+    user: userSlice,
+    post: postSlice,
+  },
+});
+
+export type RootState = ReturnType<typeof store.getState>;
+export type AppDispatch = typeof store.dispatch;
```
Dennull, line 13: Does this get used anywhere?

src/App.tsx
```diff
@@ -10,13 +10,18 @@ import Register from "./pages/Register";
...
+  const dispatch = useDispatch();
 
   useEffect(() => {
```
Dennull, line 23: We should look into if we can use `createAsyncThunk` to do this initial fetch (which comes installed with Redux Toolkit)

src/App.tsx
```diff
@@ -10,13 +10,18 @@ import Register from "./pages/Register";
...
 import "./App.css";
+import { RootState } from "./state/store";
+import { useDispatch, useSelector } from "react-redux";
+import { setUser } from "./state/slices/userSlice";
 import { PostContextProvider } from "./context/PostContext";
 
 function App() {
   const { isLoading, isAuthenticated, getAccessTokenSilently } = useAuth0();
-  const { user, setUser } = useUserContext();
+  // const { user, setUser } = useUserContext();
```
Dennull: Some comments we don't need, let's clean those up. There are some in `PostList.tsx` and `Register.tsx` too

Dennull (commented):
I had a few suggestions to clean some things up, but great job on this!

src/components/Posts/PostItem.tsx
Dennull: We'll create a bigger ticket around general error handling and tackle this there. An idea was to do undefined checks higher up in the component tree and redirect users to another route if their `userId` is undefined

src/App.tsx
Dennull: We'll also handle this in another ticket. Looks like `createAsyncThunk` just adds some syntax sugar to make things more neat

Dennull (approved):

Issue #29, opened 2023-04-11
Fix post actions numbers


Dennull:
STEPS:

1. Go to any post
2. Like or unlike the post

OBSERVED:

- Sometimes all of the other icons in the post actions area will shift over when the number of likes changes

EXPECTED:

- The number of likes changing shouldn't affect the other icons in the post actions area

AnindaS13:
I added a position property with value of absolute to actionitems style in PostItem.tsx. Changed padding since that made the numbers misplace and don't see the issue anymore. The branch is linked to this ticket.

Dennull:
Nice, can you make a PR once it's ready?

PR #41, opened 2023-05-25
T20: Expanded post page & ComposePost functionality

### LINKED ISSUE #20: Expanded post view
This is the page that users see when they click on a specific post on their timeline or someone's profile. It includes an expanded post view and replies to the post. Refer to the Figma for what the UI looks like. MVP requirements:

- Navigate to a route which will display the post and replies
- The component will retrieve data from the post to render initially from the Redux store. It will then make a fetch for the post information in case the user directly navigated to the post using a link or the post data was changed
- Users should be able to like posts and replies
- Users should be able to reply using a text box at the top of the list of replies
  - Clicking on the reply button will open a modal, but this will be done in a different ticket

NasihNazeem:
This PR contains the introduction of the Expanded Post page where you are able to:

1. Click on any post, to view post details and replies
2. Reply to any post
3. Like replies and selected post
4. Navigate to a previous page from Expanded Post
5. Link directly onto any post, and gather all details of the page.

There are critical changes in terms of styling on ExpandedPost and functionality for ComposePost and ComposeReply.

- ComposePost and Reply now use the Redux store to append any new post/reply into the list of Posts in the store.
- Reply is a different component due to some functionality changes with regards to how a reply component should operate and how the CompostPost should operate (and look).


AnindaS13:
When I like a reply to a reply of a post I get this pop up on the like count:

image

AnindaS13:
imageimage

When I like and unlike a post, the number being changed causes a shift in the place of the part I highlighted. Similar to that bug we had before, but I don't remember how we fixed it: https://github.com/TweetClone/twitter-clone-frontend/issues/29

Dennull:
Should we just change the behavior of `ComposePost` to be the same as `ComposeReply`? That way we might be able combine both into one component

Maybe we want the action buttons on `ComposePost` to always be visible since we want it to take up more space though. I guess the point of hiding the action buttons on `ComposeReply` was to make it easier to differentiate between your timeline and a list of replies. Let me know what you guys think

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+    navigate(path, { state: post.postId });
+  };
+
+  return (
+    <Card sx={styles.card}>
+      <Box style={styles.topHeader}>
+        <IconButton style={styles.backButton} onClick={() => navigate(-1)}>
+          <KeyboardBackspaceIcon color="secondary" />
+        </IconButton>
+        <Typography style={styles.headerTitle}>Tweet</Typography>
```
Dennull: Let's change this to be "Post" instead of "Tweet" so we're not completely ripping Twitter off lol

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+      </CardActionArea>
+
+      <Box sx={styles.actionData}>
+        <Typography component={"span"} sx={styles.timestamp}>
+          {post.timestamp}
+        </Typography>
+      </Box>
+      <Divider variant="middle" />
+      <Box sx={styles.actionData}>
+        <Typography component={"span"} sx={styles.actionNumbers}>
```
Dennull: 2 things here:

1. I think the reason why the bug where the numbers shift that Aninda pointed out is happening is because you nested a `Typography` component inside another `Typography` component
2. Can we replace this with a button so that in the future we can show who's like/reposted something on click?

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+        avatar={<Avatar>CK</Avatar>}
+        action={
+          <IconButton>
+            <MoreVertIcon />
+          </IconButton>
+        }
+        title={`${post.displayName} `}
+        subheader={`@${post.username}`}
+      />
+      <CardActionArea onClick={() => routeChange()}>
```
Dennull: Can we remove the onClick on this? Clicking on an expanded post shouldn't redirect you anywhere. Since that's the case let's get rid of the `CardActionArea` too, and just put the post content in `CardContent`

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+  },
+};
+
+const ExpandedPostItem = () => {
+  const dispatch = useAppDispatch();
+  const user = useAppSelector((state) => state.user);
+  const post = useAppSelector((state) => state.post);
+  const urlParams = useParams();
+
+  // TODO: fetch post data when a user may redirect directly to this link
```
Dennull: If this comment's been done already let's delete it

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -0,0 +1,224 @@
...
+        </Typography>
+      </Box>
+      <Divider variant="middle" />
+      <CardActions>
+        <Stack
+          direction="row"
+          justifyContent="space-between"
+          sx={styles.cardActions}
+        >
+          <Button
```
Dennull: Small nitpick, but since we aren't displaying numbers in these buttons I think we can use MUI's `IconButton` instead of `Button`. I don't think it makes much of a difference but the styling looks a little better (it's round instead of square) and just a more proper use case

src/components/Posts/PostList.tsx
```diff
@@ -2,24 +2,49 @@ import { useEffect } from "react";
...
   useEffect(() => {
     const fetchPosts = async () => {
-      const result = await axios.get("http://localhost:3001/api/posts", {
-        params: {
-          userId: user.userId,
-        },
-      });
-      dispatch(setPosts(result.data as Post[]));
+      try {
+        if (replies) {
```
Dennull: I think it's worth looking into how we can replace this `replies` prop with something better. In #42, we have 4 more post list components that have the same issue since all of our post lists are basically the same other than which endpoint they fetch data from

I feel like there must be a better way to determine which endpoint to use. Let's loop in @AnindaS13 too since he did the profile page post lists. Maybe we can let this slide in this PR and make a new ticket to figure this stuff out

src/pages/Profile.tsx
```diff
@@ -1,5 +1,19 @@
...
+
+const styles = {
+  root: {
+    width: "100%",
+  },
+  postListContainer: { padding: 2, width: "50%" },
+};
+
 const Profile = () => {
-  return <h1 style={{ margin: "auto" }}>Profile</h1>;
```
Dennull: Since this is getting overhauled in another PR I think we should remove these changes and leave this page the way it was before

src/state/slices/expandedPostSlice.ts
```diff
@@ -0,0 +1,29 @@
...
+  textContent: "",
+  timestamp: "",
+  username: "",
+};
+
+export const postSlice = createSlice({
+  name: "post",
+  initialState,
+  reducers: {
+    setPost: (_, action: PayloadAction<Post>) => {
```
Dennull: Can we please rename this reducer to "setExpandedPost"?

src/state/slices/postsSlice.ts
```diff
@@ -29,9 +29,13 @@ export const postSlice = createSlice({
       });
       return newPosts;
     },
+    appendPost: (state, action: PayloadAction<Post>) => {
+      state.push(action.payload);
```
Dennull: This appends to the end of the array, but I think we want to append new posts to the front of the array so they appear first on the user's timeline. We should use something like `.unshift()` instead, which appends to the front of an array

We'll take a performance hit for sure since appending to the front of an array means you have to copy all of the items in the array back again, but I can't think of a way around it. It'd be super annoying to have to scroll to the bottom of your feed to see your new post, so I think I'm fine with taking the performance nerf

Dennull:
Another bug is that the `ComposeReply` component's action items don't re-hide if the user navigates from the expanded post to a reply. Here's some steps to reproduce:

1. Open an expanded post
2. Click on the reply text field
3. Click on one of the replies

After step 2, the action items below the reply text field pop up like expected. But when you click on a reply, the action items are still there. The action items should re-hide and then pop up only after you focus on the text field again

Dennull (changes requested):
Great job, I like it! I requested a few changes but the core functionality is there

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: Fixed this issue, thanks

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: Updated this

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: This has been updated

src/components/Posts/PostList.tsx
NasihNazeem: I agree with this, there needs to be a better way for us to use the PostList template. Seems like a lot of copy paste right now.

I'll create a temporary page called ExpandedPostList just to remove the replies prop, but we'll need to create an issue to replace all of these components.

src/pages/Profile.tsx
NasihNazeem: Not sure where I got this change from, will revert it

src/state/slices/expandedPostSlice.ts
NasihNazeem: This has been updated

src/state/slices/postsSlice.ts
NasihNazeem: Agree, if there are any major performance spikes then we can look at it at a later date.

For now, this has been implemented, but we would also have to reverse the array in the fetch query for posts. This can probably be fixed in the query itself by using something like `ORDER BY timestamp DESC`

src/components/Posts/PostList.tsx
Dennull: Yeah, I'm down for that. I make a new ticket to handle refactoring these post lists: #45

src/state/slices/postsSlice.ts
Dennull: Nice! And yeah you're right, we'll have to fix the query to use `ORDER BY timestamp DESC`. I think Aninda threw that in one of his profile page PRs so I was thinking we could merge it in at the same time. If not I'll make a new ticket for it

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -0,0 +1,230 @@
...
+        />
+      )}
+      <Box sx={styles.timestampBox}>
+        <Typography component={"span"} sx={styles.timestamp}>
+          {post.timestamp}
+        </Typography>
+      </Box>
+      <Divider variant="middle" />
+      <Box sx={styles.actionData}>
+        <Button component={"span"} sx={styles.actionNumbers}>
```
Dennull: I have a few points:

1. Was there a reason to make the button a span? Can we remove that and make it a normal button? Or a div?
2. Since the number and action are still in different containers, we're still getting that bug where the numbers shift when they change. I think the fix is to do something like:
```
<Button>
  <Typography>{`${post.numberOfLikes} Likes`}</Typography>
</Button>
```

src/components/Posts/ComposeReply.tsx
```diff
@@ -0,0 +1,132 @@
...
+              hiddenLabel
+              id="standard-multiline-static"
+              multiline
+              onChange={(e) => setPostTextContent(e.target.value)}
+              placeholder={placeholder}
+              sx={styles.textField}
+              value={postTextContent}
+              variant="standard"
+              onFocus={() => setFocusReply(true)}
+              onBlur={() => setFocusReply(false)}
```
Dennull: Adding an `onBlur` still makes the component act weird, cause when you try to click on the icons they disappear right away. This happens because the `onBlur` is only on the `TextField`, and the icons are outside of the `TextField`

I'm not really sure how to solve at first glance. Wish we could make the `onBlur` on the parent container instead.. If you have trouble coming up with something we can pair program and try stuff together

Dennull (changes requested):

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: 1. Tbh I can't remember why I did that, but it works without it so I removed it
2. The update made was so that it doesn't shift the entire row, but if we're talking about the shift within the "Likes" container for example I don't think that solves it. I tried to fix it with this commit: https://github.com/TweetClone/twitter-clone-frontend/pull/41/commits/3fa8a6fd82e163e2900d8720bcdd246c25024fff

It's not a complete fix cuz it looks bad but it does get rid of the glitch

Honestly, the issue persists unless we have a `margin: auto` or something

src/components/Posts/ComposeReply.tsx
NasihNazeem: Yeah this was my temporary solution tbh, I don't like it either. This was more of a focus to remove the icons when you click off, but I guess that would take away from the icon functionality either 

I can't really update it because like you said I don't have access to it since the action is in a separate component to the Compose Reply component

Dennull:
> imageimage
> 
> When I like and unlike a post, the number being changed causes a shift in the place of the part I highlighted. Similar to that bug we had before, but I don't remember how we fixed it: #29

Let's keep this the way it is for now and we'll make a new ticket to handle it. I have a feeling that it's something wrong with the markup, not the css, but we'll look into it some other time

src/components/Posts/ExpandedPostItem.tsx
Dennull: As discussed, we'll revert this to the way you originally had it and make a new ticket to address this bug

src/components/Posts/ComposeReply.tsx
Dennull: As discussed, we'll revert this to the way you originally had it and make a new ticket to more fully flesh out the behavior of the buttons

src/components/Posts/ComposePost.tsx
```diff
@@ -24,21 +31,28 @@ const styles = {
...
       });
       setPostTextContent("");
-      // TODO: Update posts in postContext so that the page rerenders rather than refreshing
-      window.location.reload();
+      dispatch(
+        appendPost({
+          ...newPost.data,
+          username: user.username,
+          displayName: user.displayName,
+          numberOfLikes: 0,
```
Dennull: I agree with this solution for the NaN thing, but can we please move this into the reducer? It'll save us from having to set the number of likes in every dispatch

src/components/Posts/ComposeReply.tsx
```diff
@@ -0,0 +1,131 @@
...
+          textContent,
+        }
+      );
+      setPostTextContent("");
+      dispatch(
+        appendPost({
+          ...reply.data,
+          username: user.username,
+          displayName: user.displayName,
+          numberOfLikes: 0,
```
Dennull: Same thing here as my comment in `ComposePost.tsx`, let's move this into the reducer

Dennull (approved):

PR #42, opened 2023-05-27
T16: Profile page

### LINKED ISSUE #16: User profile page
This is the user profile which contains their avatar, background photo, bio, etc. The user's posts will render the timeline post component used on the timeline. MVP requirements:

- User should see their profile details excluding pictures and follower/following counts (username, display name, join date, etc)
  - Hard code a banner and either hard code or exclude follower counts for now. We'll address these in a different ticket
- Component should have tabs to filter between all of the user's posts, their replies, and their liked posts

AnindaS13:
This PR contains the complete profile page UI. The media tab was removed and the follower/following counts have been hardcoded as discussed. All three tabs (tweets, replies, likes) are functional with the backend ticket branch.


src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+import EditIcon from "@mui/icons-material/Edit";
+import ProfileTweets from "../components/Posts/ProfileTweets";
+import axios from "axios";
+import { useAppSelector } from "../state/hooks";
+import ProfileReplies from "../components/Posts/ProfileReplies";
+import ProfileLikes from "../components/Posts/ProfileLikes";
+import CalendarMonthIcon from "@mui/icons-material/CalendarMonth";
+
+const styles = {
+  userProfileArrow: {},
```
NasihNazeem: Can we remove `userProfileArrow` and `rightBound` since they are not doing anything?

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+  const [bio, setBio] = React.useState();
+  const [joinDate, setJoinDate] = React.useState("YYYY-MM-DD");
+  const user = useAppSelector((state) => state.user);
+  const handleChange = (event: React.SyntheticEvent, newValue: string) => {
+    setValue(newValue);
+    console.log(value);
+  };
+
+  useEffect(() => {
+    const fetchTweetCount = async () => {
```
NasihNazeem: Since all of these fetches require the same parameter, is there any way for us to manipulate the query so that it returns all the required information in just one fetch?

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+      );
+      setJoinDate(result.data);
+    };
+    fetchBio();
+    fetchTweetCount();
+    fetchJoinDate();
+  }, [value, user]);
+  return (
+    <>
+      <Box width={650}>
```
NasihNazeem: Even though we can use the `width` attribute directly on `Box`, I think we should create a separate `styles.parentBox` or something for the CSS.

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+        <div style={styles.rightBound}>
```
NasihNazeem: This might be nit-picky but I think it's best standard to follow MUI even for `div`. Basically, where you write `div`, make it `Box`.

https://stackoverflow.com/questions/60451957/what-is-the-material-ui-box-component-for

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+          {/**Not sure about this */}
+          <div style={styles.header}>
+            {" "}
+            <Typography sx={styles.arrowText}>{user.username}</Typography>
+            <button style={styles.arrowBtn}>
```
NasihNazeem: Similar to the previous comment about `Box`, let's make this an `IconButton`, provide the same styling.

I would also suggest an `onClick` where you are able to revert to the previous page. You will have to 
 ```suggestion
 //outside of the render component: 
 const navigate = useNavigate();
 
 //inside of the render component:
<IconButton style={styles.arrowBtn} onClick={() => navigate(-1)}>
```

File: src/pages/Profile.tsx [hunk 6]
```diff
@@ -1,5 +1,222 @@
...
+              <ArrowBackIcon></ArrowBackIcon>
+            </button>
+            <Typography sx={styles.tweetCountStyle}>
+              {tweetCount} Tweets
+            </Typography>
+          </div>
+          <div style={{ position: "relative" }}>
```
NasihNazeem, line 148: Remove the in-line styling and place it within the `styles` object

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+            <img
```
NasihNazeem: I believe this can also be a `Box` component with a variant of `img`. 

Take a look at this: https://stackoverflow.com/questions/61263669/does-material-ui-have-an-image-component

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+            />
+          </div>
+          <Button
+            variant="contained"
+            startIcon={<EditIcon />}
+            sx={styles.editBtn}
+          >
+            Edit Profile
+          </Button>
+          <div style={{ paddingLeft: 15, overflowWrap: "break-word" }}>
```
NasihNazeem: Move the in-line styling to `styles` object

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+            <Typography variant={"h1"} style={styles.displayName}>
+              {user.displayName}
+            </Typography>
+            <Typography variant={"h2"} style={styles.usernameDisplay}>
+              @{user.username}
+            </Typography>
+            <Typography style={styles.bioDisplay}>{bio}</Typography>
+            <div style={{ alignItems: "center", display: "flex" }}>
```
NasihNazeem: Move the in-line styling to `styles` object

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+              <CalendarMonthIcon></CalendarMonthIcon>
+              <Typography>Joined {joinDate}</Typography>
+            </div>
+            <Button sx={styles.followercountbtn}>
+              <div>
+                <b>500M</b> Followers
```
NasihNazeem, line 189: Try to use `Typography` for this

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+              </div>
+            </Button>
+            <Button sx={styles.followingcountbtn}>
+              <div>
+                <b>0</b> Following
```
NasihNazeem, line 194: Same comment as above

NasihNazeem (changes requested):

src/components/Posts/ProfileLikes.tsx
NasihNazeem, line 1: I like the simplicity, and using the PostList as a template. Is there any way for us to combine `ProfileLikes`, `ProfileReplies` and `ProfileTweets` or do you think that would make the component too complex?

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+              </div>
+            </Button>
+          </div>
+          <Tabs value={value} onChange={handleChange}>
```
NasihNazeem: Good use of Tabs component!

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+            <Tab value="one" style={styles.tabSelection} label="Tweets" />
+            <Tab value="two" style={styles.tabSelection} label="Replies" />
+            <Tab value="three" style={styles.tabSelection} label="Likes" />
+          </Tabs>
+
+          {value === "one" && (
+            <div>
+              {" "}
```
NasihNazeem: What does the empty brackets do?

src/components/Posts/ProfileLikes.tsx
Dennull: I agree with this. The post list components are basically the same except for the endpoint they fetch data from. I commented this on #41, but we have the same issue in our reply list for an expanded post. I suggested on there that we could make a new ticket to sort that out

src/pages/Profile.tsx
Dennull: Seconded, it's a nitpick but for consistency I think we should stick to keeping our css in the `styles` object unless there's a good reason for inlining it

src/pages/Profile.tsx
Dennull: Yeah, was actually going to bring this up at a meeting, but we should talk about how we feel about using MUI's component over normal html elements. I don't have a strong opinion but it's probably better to be consistent in the codebase

I had similar question with MUI's `Stack` component, which is just a flexbox container. You can use a `div` or `Box` and put flex properties on it to achieve the same thing, so I got confused about which one I should use

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
+        <div style={styles.rightBound}>
+          {/**Not sure about this */}
+          <div style={styles.header}>
+            {" "}
+            <Typography sx={styles.arrowText}>{user.username}</Typography>
+            <button style={styles.arrowBtn}>
+              <ArrowBackIcon></ArrowBackIcon>
+            </button>
+            <Typography sx={styles.tweetCountStyle}>
+              {tweetCount} Tweets
```
Dennull: In terms of styling my preference is that the header has bigger font than the number of tweets. Let me know if anyone has a feels differently

src/pages/Profile.tsx
```diff
@@ -1,5 +1,222 @@
...
 const Profile = () => {
-  return <h1 style={{ margin: "auto" }}>Profile</h1>;
+  const [value, setValue] = React.useState("one");
+  const [tweetCount, setTweetCount] = React.useState();
+  const [bio, setBio] = React.useState();
+  const [joinDate, setJoinDate] = React.useState("YYYY-MM-DD");
+  const user = useAppSelector((state) => state.user);
+  const handleChange = (event: React.SyntheticEvent, newValue: string) => {
+    setValue(newValue);
+    console.log(value);
```
Dennull: Can we please remove this log?

src/pages/Profile.tsx
Dennull: One nitpick, I think since `handleChange` isn't being reused and its logic isn't overly complex, we can inline it as an arrow function. My suggestion is to do something like this instead:

```suggestion
          <Tabs value={value} onChange={(_: React.SyntheticEvent, newValue: string) => setValue(newValue)}>
```

I prefer inlining in this case because it's not a lot of code. If the onChange was a lot more code I'd agree with this approach to not bloat the jsx

src/pages/Profile.tsx
Dennull: There are a few of these in this file, can we please remove them?

Dennull (changes requested):

src/pages/Profile.tsx
```diff
@@ -1,5 +1,219 @@
...
+    display: "flex",
+  },
+};
+
 const Profile = () => {
-  return <h1 style={{ margin: "auto" }}>Profile</h1>;
+  const navigate = useNavigate();
+  const [value, setValue] = React.useState("one");
+  const [profileContents, setProfileContents] = React.useState({
+    tweetcount: "",
```
Dennull: Can we please rename these to be camel case? Same goes for the backend the returns them

src/pages/Profile.tsx
```diff
@@ -1,5 +1,219 @@
...
+      setProfileContents(result.data);
+    };
+
+    fetchProfileContents();
+  }, [value, user]);
+  return (
+    <>
+      <Box style={styles.parentBox}>
+        <Box>
+          {/**Not sure about this */}
```
Dennull: Let's remove this comment if we don't need it

src/pages/Profile.tsx
```diff
@@ -1,5 +1,219 @@
...
+              {user.displayName}
+            </Typography>
+            <Typography variant={"h3"} style={styles.usernameDisplay}>
+              @{user.username}
+            </Typography>
+            <Typography style={styles.bioDisplay}>
+              {profileContents.bio}
+            </Typography>
+            <Box style={styles.calendarBox}>
+              <CalendarMonthIcon></CalendarMonthIcon>
```
Dennull: Can we please make components without children self closing?
```suggestion
              <CalendarMonthIcon/>
```

Dennull (approved):
I have a few nitpicks and ideas about how we can simplify some markup, but I don't think any of them are worth blocking the PR over. If you're down tho we can get on call one day before merging and talk about it!

src/pages/Profile.tsx
NasihNazeem: agree

NasihNazeem (approved):

PR #43, opened 2023-05-28
T12: Messages list

### LINKED ISSUE #12: Conversation list
This displays the list of conversations a user has with other users. MVP requirements:

- Users should be able to click on the "messages" tab on the navbar and see a list of their current conversations
  - The conversation list should include the other user's display name, username, the date of the latest message, and the content of the latest message
- The right side of the page where the DM should show a default page asking the user to select a conversation if they haven't clicked on any yet

Dennull:
Closes #12. This is the list of conversations a user has with other people. The direct message will be done in a separate PR. There are also a few styling inconsistencies I'd like to clean up too, but I'll make another ticket to tackle those since I think they apply to more than just this page


src/components/Messages/SearchBar.tsx
```diff
@@ -0,0 +1,37 @@
...
+  },
+};
+
+const SearchBar = () => {
+  return (
+    <Box sx={styles.box}>
+      <TextField
+        fullWidth
+        hiddenLabel
+        InputProps={{
```
NasihNazeem, line 26: I like this!

src/components/Messages/MessagesListItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const MessagesListItem = ({ latestMessageDetails }: MessageListItemProps) => {
+  return (
+    <ListItemButton>
+      <Stack direction="row">
+        <ListItemAvatar sx={styles.avatar}>
+          <Avatar />
+        </ListItemAvatar>
+        <ListItemText
+          primary={`${latestMessageDetails.displayName} @${latestMessageDetails.username} - ${latestMessageDetails.timestamp}`}
```
NasihNazeem: This might be something we fix later, but I think we should figure out a way for us to ensure that the name does not spill over to a second line. Right now the timestamp spills over, which will be fixed at a later date, but if we end up creating a long display name this is something that should be cut-off by maybe adding `"..."` or something at the end of the string. 

That way the Items don't lose that clean look.

NasihNazeem: I'm also not sure how we can do this with the `primary` attribute, but is there any way we can change the color of the username so that it doesn't mesh with the display name?

src/components/Messages/SearchBar.tsx
NasihNazeem, line 1: I like this component! I wonder if we can make the component more reusable by adding a `placeholder` prop similar to `ComposePost` where we can implement this as the SearchBar in the `Timeline` page

src/pages/Messages.tsx
```diff
@@ -1,5 +1,68 @@
...
+  },
+};
+
 const Messages = () => {
-  return <h1 style={{ margin: "auto" }}>Messages</h1>;
+  return (
+    <Stack
+      direction="row"
+      sx={styles.root}
+      divider={<Divider orientation="vertical" />}
```
NasihNazeem, line 40: Nice use of the divider attribute, I'd like to implement the `Stack` as the parent element in other pages as well.

src/pages/Messages.tsx
```diff
@@ -1,5 +1,68 @@
...
+      <Box sx={styles.messageListContainer}>
+        <Box sx={styles.messagesHeader}>
+          <Typography variant="h6">Messages</Typography>
+          <IconButton>
+            <ChatIcon />
+          </IconButton>
+        </Box>
+        <SearchBar />
+        <List component="div">
+          <Divider />
```
NasihNazeem: Do we need the `List` component here or can we implement something like `Box` again? I initially read this as we have a list of `Divider` and a `MessagesList` within a List. 

Let me know your thoughts

src/pages/Messages.tsx
```diff
@@ -1,5 +1,68 @@
...
+          <MessagesList />
+        </List>
+      </Box>
+      <Box sx={styles.selectMessageContainer}>
```
NasihNazeem, line 53: Bit of a nit-pick but I want to suggest having the same amount of whitespace on the right side as we do on the left side of the `NavBar`. 

More of a design choice and if it looks weird we can ignore this, let me know what you guys think

src/pages/Messages.tsx
```diff
@@ -1,5 +1,68 @@
...
+        <Typography variant="h6">Select a Message</Typography>
+        <Typography>
+          Choose one of your existing conversations or start a new one!
+        </Typography>
+        <Button variant="contained" sx={styles.button}>
+          New Message
+        </Button>
+      </Box>
```
NasihNazeem, line 61: I think we should add a divider on the other side of the Messages Box, but that would mean we need to make changes elsewhere (in the `styles.selectMessageContainer`). 

I tried it out, and the changes we would need to make it fixing the margin left and right to a certain amount of pixels (`marginLeft: "40px"` , `marginRight: "40px"`) and also decreasing the `width` to `40%`.

If you agree with the design change, but wanna do it in a different way, feel free to!

```suggestion
      </Box>
      <Divider  orientation="vertical" />
```

NasihNazeem (changes requested):

NasihNazeem (approved):

PR #47, opened 2023-06-05
T44: Create message modal

### LINKED ISSUE #44: New message modal
Upon clicking a "new message" button, a modal should pop up that allows users to search up other users who they can start a new conversation with

The implementation should follow a similar pattern to #36

NasihNazeem:
This feature allows users to create a new message to another user through a CreateMessage button on the `Messages.tsx` page.

This modal provides a list of users the active user may have recently messaged for quick action, and a search bar for the active user to search for any users they currently follow. The constraint with this search bar is that it only displays people that the active user has followed.

Once you click on the search icon, the modal routes you to a messages page where you can begin your conversation with the selected user. The actual messaging functionality is out of scope of this PR.

**TECH DEBT:**
- Currently, the modal has its own `List` and `ListItem` components, those should be updated to create something more universal where we can input JSX as a prop from a parent component.
 
- In the search bar the `Autocomplete` component has a custom `renderOption` attribute where we render a `ListItemText` element that requires a `primary` and `secondary` attribute. These take in elements themselves, but this should be updated so that we can pass this as a prop from a parent component as well.

- Figure out how to change the layout of the dropdown list (the `Autocomplete` options) so that it fills the width of the modal and doesn't look like a "dropdown". Ideally the options should look exactly like the "recent messages" list under the search bar


AnindaS13 (changes requested):
image

I feel like theres a bit too much whitespace between the magnifying glass and the text beside it. What do you think of moving the text in the search bar over to align with New Message vertically? That might reduce enough whitespace. Or do you think that makes it look weirder. Let me know.

AnindaS13:
d

AnindaS13:
image

I feel like it kinda looks weird with both the displayname and the @username beside each other like that when you select the result in the search bar. The dropdown of peoples names is fine (and probably needed to differentiate two ppl with the same display name). Do you guys think that just having the username in the search tab once you select it would look a little less awkward?

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+    display: "flex",
+    width: "100%",
+    alignItems: "center",
+    justifyContent: "space-between",
+  },
+};
+
+type CreateMessageModalProps = {
+  onClose: () => void;
+  openModal: boolean;
```
Dennull: Nitpick, can this be called open?

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+              onClick={() => routeChange()}
+              sx={styles.modalButton}
+            >
+              Message
+            </Button>
+          </Box>
+        </Box>
+      </DialogTitle>
+      <SearchBarMessages
+        placeholder="Start a conversation"
```
Dennull: Can this say "Search followers" or something similar? Since we're restricting it to that condition I think that makes more sense

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+        selectedUser={selectedUser}
+        setSelectedUser={(state) =>
+          state
+            ? setSelectedUser(state)
+            : setSelectedUser({
+                displayName: "",
+                otherUserId: -1,
+                username: "",
+              })
+        }
```
Dennull: Maybe it won't matter since we might get rid of this state, but it kinda bothers me how this looks. I think we can move the ternary down into the child component rather than putting it here:

```suggestion
        setSelectedUser={(state) => setSelectedUser(state)}
```

And then in the child component do something like this:

```
// SearchBarMessages.tsx
onChange={(_, value) =>
 value ? setSelectedUser(value) : setSelectedUser({ displayName: "" ... })
```

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+  onClose: () => void;
+  openModal: boolean;
+};
+
+export default function CreateMessageModal({
+  onClose,
+  openModal,
+}: CreateMessageModalProps) {
+  const user = useAppSelector((state) => state.user);
+  const [selectedUser, setSelectedUser] = useState<OtherUser>({
```
Dennull: I think there's some overlap here with what I have in #54. I made a new Redux slice with a similar state, so I'd say let's wait until that PR gets merged before we continue with this one cause we might be able to share that store

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+      <DialogTitle sx={styles.dialogTitle}>
+        <IconButton onClick={onClose}>
+          <CloseIcon />
+        </IconButton>
+        <Box sx={styles.modalHeader}>
+          <Box sx={styles.titleBox}>
+            <Typography sx={styles.headerTitle}>New Message</Typography>
+          </Box>
+          <Box sx={styles.modalButtonBox}>
+            <Button
```
Dennull: This button seems a little redundant to me in practice. Since we can't add multiple people into a group chat like other apps can, it makes me question if we even need it

I also don't like how we need to click this button to proceed when using the search bar, but we don't need to click it when clicking on the quick select list below. The pattern is just inconsistent. I'd suggest we either ditch the button all together or make the user click this button to proceed in both cases

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+            ? setSelectedUser(state)
+            : setSelectedUser({
+                displayName: "",
+                otherUserId: -1,
+                username: "",
+              })
+        }
+        setFocusSearchBar={(state) =>
+          state ? setFocusSearchBar(state) : setFocusSearchBar(false)
+        }
```
Dennull: Similar story here as with the other state setter, I think I'd prefer if we try to mirror the `onOpen` and `onClose` handlers than do this. So in the child component rather than take in a `setFocusSearchBar` prop, we'd take and `onOpen` and `onClose` prop

```suggestion
        onClose={() => setFocusSearchBar(false)}
        onOpen={() =>setFocusSearchBar(true)}
```

src/components/Messages/CreateMessageModal/MessagesModalListItem.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const user = useAppSelector((state) => state.user);
+  const navigate = useNavigate();
+  const routeChange = () => {
+    const path = `/messages/${user.userId}/${otherUser.otherUserId}`;
+    navigate(path);
+  };
+
+  return (
+    <ListItemButton onClick={() => routeChange()}>
+      <Stack direction="row" sx={styles.stack}>
```
Dennull: Can we use a `Box` here to follow our pattern? Let's only use `Stack` if there's a specific reason to

src/components/Messages/CreateMessageModal/SearchBarMessages.tsx
```diff
@@ -0,0 +1,157 @@
...
+  },
+  primaryTextContainer: {
+    gap: 0.5,
+    width: "30%",
+    height: "100%",
+  },
+  stack: {
+    width: "100%",
+  },
+  autocomplete: { "	.MuiAutocomplete-input": { paddingLeft: 0 } },
```
Dennull: Can you please remove the extra spaces? Also I think you need the "&" in front of the period right?

src/components/Messages/CreateMessageModal/SearchBarMessages.tsx
```diff
@@ -0,0 +1,157 @@
...
+          return (
+            <TextField
+              {...params}
+              fullWidth
+              hiddenLabel
+              InputProps={{
+                ...params.InputProps,
+                startAdornment: (
+                  <InputAdornment position="start">
+                    <IconButton
```
Dennull: Cool!

src/components/Messages/CreateMessageModal/SearchBarMessages.tsx
```diff
@@ -0,0 +1,157 @@
...
+              placeholder={placeholder}
+              size="small"
+              sx={styles.searchField}
+            />
+          );
+        }}
+        renderOption={(props, option) => {
+          return (
+            <List component="li" {...props}>
+              <Stack direction="row" sx={styles.stack}>
```
Dennull: Same here, can we use a `Box` instead?

src/pages/Messages.tsx
```diff
@@ -33,6 +35,8 @@ const styles = {
 };
 
 const Messages = () => {
+  const [messageModal, setMessageModal] = useState(false);
```
Dennull: Nitpick, what do you think about changing this to "showMessageModal"? I think it's a bit more descriptive

Dennull (changes requested):
Nice work! I think we should wait until I merge in #54 before we merge this since there's some stuff there that we might be able to use here too

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
NasihNazeem: I will update this to say "Search following" because it's people you followed rather than the people that follow you

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
NasihNazeem: updated this, thanks!

For some reason I had it both in the child and parent, but I changed it to your suggestion

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
NasihNazeem: I can remove the button since we're not going to be doing anything like multiple people in a group

src/components/Messages/ConversationList.tsx
```diff
@@ -50,16 +53,16 @@ const ConversationList = () => {
...
         </IconButton>
       </Box>
       <SearchBar placeholder="Search Messages" />
       <Divider />
       <List component="div">
-        {conversations.map((o) => (
+        {conversations.map((o, index) => (
           <ConversationListItem
-            key={o.otherUserId}
+            key={`${o.otherUserId}-${index}`}
```
Dennull: Do we need the index here?

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+} from "@mui/material/";
+import CloseIcon from "@mui/icons-material/Close";
+import SearchBarMessages from "./SearchBarMessages";
+import MessagesModalList from "./MessagesModalList";
+import { useState } from "react";
+
+const styles = {
+  dialog: {
+    height: "60%",
+    borderRadius: 4,
```
Dennull: That's weird, I thought this was taken care of in `Theme.tsx`. Can you take a look to see why it doesn't work? Our other dialogs have rounded borders without having to set the border radius, so idk why this one isn't the same

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+  },
+  dialogContent: {
+    padding: 0,
+  },
+  dialogTitle: {
+    paddingBottom: 0,
+    paddingLeft: 0.5,
+    paddingRight: 0.5,
```
Dennull: Nit, can use `paddingX` to save some lines

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+    paddingTop: 1,
+    display: "flex",
+    alignItems: "center",
+  },
+  titleBox: { paddingLeft: 3, width: "100%" },
+  headerTitle: {
+    fontWeight: "bold",
+  },
+  modalButtonBox: { paddingRight: 2 },
+  modalButton: { borderRadius: 10 },
```
Dennull: I don't see these used right?

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
```diff
@@ -0,0 +1,85 @@
...
+      onClose={onClose}
+      open={open}
+      scroll="paper"
+      PaperProps={{ sx: styles.dialog }}
+    >
+      <DialogTitle sx={styles.dialogTitle}>
+        <IconButton onClick={onClose}>
+          <CloseIcon />
+        </IconButton>
+        <Box sx={styles.modalHeader}>
```
Dennull: Does this div do anything? If not can we remove it and the corresponding style?

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
```diff
@@ -0,0 +1,45 @@
+import axios from "axios";
+import { useState, useEffect } from "react";
+import { List } from "@mui/material";
+import { useAppSelector } from "../../../state/hooks";
+import MessagesModalListItem from "./MessagesModalListItem";
+
+export type OtherUser = {
```
Dennull: I think I said I'd leave it up to you to decide what to do with this type, did you decide to keep it? Just curious

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
```diff
@@ -0,0 +1,45 @@
...
+      );
+      setConversationList(result.data as OtherUser[]);
+    };
+    fetchConversationList();
+  }, [user]);
+
+  return (
+    <List component="div">
+      {conversationList.map((o, index) => (
+        <MessagesModalListItem otherUser={o} key={index} onClose={onClose} />
```
Dennull: Do you think we can use the other user's id as the key? You can only have one conversation with each unique user, so I think it makes sense

src/components/Messages/CreateMessageModal/MessagesModalListItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const navigate = useNavigate();
+  const routeChange = () => {
+    onClose();
+    const path = `/messages/${user.userId}/${otherUser.otherUserId}`;
+    navigate(path);
+  };
+
+  return (
+    <ListItemButton onClick={() => routeChange()}>
+      <Box sx={styles.stack}>
```
Dennull: Nitpick, this isn't a stack anymore, so can we change the class name?

src/components/Messages/CreateMessageModal/MessagesModalListItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+        <ListItemAvatar sx={styles.avatar}>
+          <Avatar />
+        </ListItemAvatar>
+        <ListItemText
+          disableTypography
+          primary={
+            <Box sx={styles.primaryTextContainer}>
```
Dennull: Does `primaryTextContainer` do anything? If not can we remove it?

src/components/Messages/CreateMessageModal/SearchBarMessages.tsx
Dennull, line 1: Can we rename this component to `MessagesSearchBar`? I think that sounds better, but let me know what you think. I also proposed some styling changes in #62. Let me know what you think about them!

src/components/Messages/CreateMessageModal/SearchBarMessages.tsx
```diff
@@ -0,0 +1,155 @@
...
+      setFollowedList(result.data as OtherUser[]);
+    };
+    fetchDMList();
+  }, [user]);
+
+  return (
+    <Box sx={styles.box}>
+      <Autocomplete
+        onOpen={() => setFocusSearchBar(true)}
+        onClose={() => setFocusSearchBar(false)}
```
Dennull: Instead of passing `setFocusSearchBar` as a prop, can we include this state change in the `onClose` function and `onOpen` function and have those as props instead? That way all `onClose` functionality is wrapped up in one function

src/pages/Messages.tsx
```diff
@@ -12,28 +14,47 @@ const styles = {
     display: "flex",
     flexDirection: "column",
     justifyContent: "center",
-    alignItems: "center",
+    alignItems: "flex-start",
+    margin: "auto",
     width: "50%",
   },
+  messagesHeader: {
```
Dennull: This isn't being used

Dennull (changes requested):
Looks mostly good, just some code clean up and a few small proposals!

src/components/Messages/ConversationList.tsx
NasihNazeem: Nope, this is outdated code.

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
NasihNazeem: I think it has to do with me having the set the height of the dialog through `sx` within `PaperProps` attribute. This seems to override the `Theme.tsx` and if I don't maintain a certain height, it just goes back to minHeight

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
NasihNazeem: Thanks, they aren't being used

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
NasihNazeem: Nope, outdated styling

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
NasihNazeem: Yeah I decided to keep it. I felt like it has a different purpose than `SelectedConversation`.

One thing we could do is rename it to something general so that it's used in both components and then apply that type individually for the required state. 

I might do that rather than have duplicate types

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
NasihNazeem: Yep this makes sense

src/components/Messages/CreateMessageModal/SearchBarMessages.tsx
NasihNazeem: So you want me to send `onOpen` and `onClose` as props to `MessagesSearchBar.tsx`? 

Would look like 
```suggestion
        onOpen={onOpen}
        onClose={onClose}
```

src/components/Messages/CreateMessageModal/MessagesModalListItem.tsx
NasihNazeem: Changed it to `container`

src/components/Messages/CreateMessageModal/SearchBarMessages.tsx
NasihNazeem: Sure thing

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
NasihNazeem: Renamed `SelectedConversation` to `SelectedUser` to fit the criteria for both `OtherUser` and `SelectedCoversation`.

src/components/Messages/CreateMessageModal/CreateMessageModal.tsx
Dennull: I see, it's fine then

src/components/Messages/CreateMessageModal/SearchBarMessages.tsx
Dennull: Resolved this offline!

Dennull (approved):

AnindaS13 (approved):

PR #51, opened 2023-06-18
T36: Reply modal

### LINKED ISSUE #36: Reply modal
Creating the reply modal that pops up and lets users reply to a post. Refer to figma for UI design

NasihNazeem:
This feature allows the user to reply to a post through a modal. When clicking on the `AddCommentOutlinedIcon` button on a Post Item, this renders a modal that takes in the respective post data, displays the original post and provides a component that allows the user to reply to that post.


src/components/NavBar/PostButtonModal.tsx
```diff
@@ -7,7 +7,10 @@ const styles = {
     ".MuiDialog-scrollPaper": { alignItems: "flex-start" },
   },
   dialogContent: {
-    padding: 0,
+    padding: 1,
```
Dennull: Do we need this stuff?

src/components/Posts/ComposeReply.tsx
```diff
@@ -71,6 +75,9 @@ const ComposeReply = ({ placeholder }: ComposeReplyProps) => {
           displayName: user.displayName,
         })
       );
+      if (onClose) {
+        onClose();
+      }
```
Dennull: I believe you can simplify this using optional chaining. Can read more about it [here](https://javascript.info/optional-chaining)

```suggestion
      onClose?.()
```

src/components/Posts/ComposeReply.tsx
```diff
@@ -59,7 +63,7 @@ const ComposeReply = ({ placeholder }: ComposeReplyProps) => {
         "http://localhost:3001/api/posts/postReply",
         {
           userId: user.userId,
-          parentPostId: post.postId,
+          parentPostId: parentPostId,
```
Dennull: Can simplify this a little since the object key is the same as the variable name

```suggestion
          parentPostId,
```

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -217,6 +227,13 @@ const ExpandedPostItem = () => {
         </Stack>
       </CardActions>
       <Divider variant="middle" />
+      {openModal && (
```
Dennull: We don't need this conditional render if we're using the `open` prop on the modal. I think we should choose either the parent or the child to control the render, not both. So we could either set `open` in the child to always be true and keep this conditional render using `openModal`, or drop the conditional render and pass `openModal` to the child

src/components/Posts/ExpandedPostReplies.tsx
```diff
@@ -32,9 +35,11 @@ const ExpandedPostReplies = () => {
 
   return (
     <>
-      {posts.map((o, index) => (
-        <PostItem key={index} post={o} />
-      ))}
+      {posts
+        .filter((o) => o.parentPostId === post.postId)
+        .map((o, index) => (
+          <PostItem key={index} post={o} />
```
Dennull: Minor thing, maybe using the `postId` as the key is a better idea since it's supposed to be our unique identifier

src/components/Posts/PostItem.tsx
```diff
@@ -141,6 +149,9 @@ const PostItem = ({ post }: PostProps) => {
             {post.numberOfLikes}
           </Button>
           <Button
+            onClick={() => {
+              replyModalPopup();
+            }}
```
Dennull: Can we get rid of the function definition and inline this instead? If the logic isn't too complicated I think it's fine to inline it

```suggestion
            onClick={() => {
              setOpenModal(true);
            }}
```

src/components/Posts/PostItem.tsx
```diff
@@ -153,6 +164,13 @@ const PostItem = ({ post }: PostProps) => {
         </Stack>
       </CardActions>
       <Divider light />
+      {openModal && (
```
Dennull: Similar story here to the comment I made in `ExpandedPostItem.tsx`, I suggest we choose either the parent or child to control the render, not both

src/components/Posts/RepliesModal.tsx
Dennull, line 1: Looks good to me overall! I'd like to see if we can simplify some of the markup and make some minor styling tweaks though. Maybe we can pair program this one day to go over the small details. Also, it looks like there's some css in the `styles` object that isn't used, can we remove those?

src/pages/ExpandedPost.tsx
```diff
@@ -2,22 +2,30 @@ import { Box, Divider } from "@mui/material";
...
-import ExpandedPostList from "../components/Posts/ExpandedPostReplies";
+import ExpandedPostReplies from "../components/Posts/ExpandedPostReplies";
+import { useAppSelector } from "../state/hooks";
 
 const styles = {
   root: {
     width: "100%",
   },
   postListContainer: { padding: 0, width: "50%" },
+  divider: { paddingBottom: 3 },
```
Dennull: Does this get used?

Dennull (changes requested):

src/components/NavBar/PostButtonModal.tsx
NasihNazeem: nah this is out of scope and random lol ima revert it

src/components/Posts/ComposeReply.tsx
NasihNazeem: Nice suggestion! I didn't know about this and was annoying me

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: okay good eye, I'll remove the conditional render and just pass it down to the child as a prop

src/components/Posts/ExpandedPostReplies.tsx
NasihNazeem: Should I make this change on `PostList.tsx` as well?

src/components/Posts/PostItem.tsx
NasihNazeem: Yeah will make this change, I think I was trying something with multiple functions but removed that part so this should have been kept in-line, thanks

src/pages/ExpandedPost.tsx
NasihNazeem: It wasn't originally used, but I ended up using it just now to help fix some padding issues on `ComposeReply.tsx` component within the reply modal which consequentially affected `ExpandedPost.tsx`

src/components/Posts/RepliesModal.tsx
NasihNazeem: Removed some of the unnecessary styles, I think we may have to do this for the entire app because I feel like I've left unused CSS on other pages as well

AnindaS13 (approved):

src/components/Posts/ExpandedPostReplies.tsx
Dennull: Yeah sure I think that'd be a nice change

Dennull (approved):

PR #53, opened 2023-06-19
T52: Icon Colours

### LINKED ISSUE #52: Post action item colors
Our post action items should have their colors changed to be grey (#adb5bd) normally and only turn green (#22aa6f) upon hover. If a user has liked or reposted a post, those buttons should change to become permanently green

Please also change the disabled color of the post button in our `ComposePost` component to dark green (#2a7946) while you're doing this ticket

NasihNazeem:
- This issue changes the display for action items within `PostItem` and `ExpandedPostItem` so that the actions are gray when inactive, green on hover, and green when active. 

    - For actions that do not have functionality yet, I have not provided the active/inactive display functionality, but this can either be done within this PR (delay the PR until comment, repost, share functionality is working on main branch) OR we add the display functionality within their respective issues.

    - For example, liking functionality works. We have a boolean that tells the user whether they liked a post or not `isLikedByCurrentUser`. For likes, the display functionality is working.

- I changed the post button disabled color from default gray to a light green on both `ComposeReply` and `ComposePost`.

- I changed the colour of the actions in `ComposePost` and `ComposeReply` to be permanently green.

### DIFF
--- a/src/components/Posts/ComposePost.tsx
+++ b/src/components/Posts/ComposePost.tsx
@@ -35,15 +35,6 @@ const styles = {
     justifyContent: "space-between",
     marginTop: 1,
   },
-  postButton: {
-    "&.Mui-disabled": {
-      backgroundColor: "primary.light",
-      color: "white",
-    },
-    borderRadius: 5,
-    height: 35,
-    weight: 35,
-  },
 };
 
 const ComposePost = ({ placeholder, minRows }: ComposePostProps) => {
@@ -90,21 +81,19 @@ const ComposePost = ({ placeholder, minRows }: ComposePostProps) => {
             sx={styles.textField}
             value={postTextContent}
             variant="standard"
-            InputProps={{ disableUnderline: true }}
           />
           <Box sx={styles.postActions}>
             <Stack direction="row">
-              <IconButton size="small" color="primary">
+              <IconButton size="small">
                 <AddPhotoAlternateOutlinedIcon />
               </IconButton>
-              <IconButton size="small" color="primary">
+              <IconButton size="small">
                 <EmojiEmotionsOutlinedIcon />
               </IconButton>
             </Stack>
             <Button
               disabled={!postTextContent.trim()}
               size="small"
-              sx={styles.postButton}
               type="submit"
               variant="contained"
             >
--- a/src/components/Posts/ComposePost.tsx
+++ b/src/components/Posts/ComposePost.tsx
@@ -28,24 +28,21 @@ const styles = {
     justifyContent: "space-between",
     marginTop: 1,
   },
-  postButton: { borderRadius: 5 },
+  postButton: {
+    "&.Mui-disabled": {
+      backgroundColor: "primary.light",
+      color: "white",
+    },
+    borderRadius: 5,
+    height: 35,
+    weight: 35,
+  },
 };
 
 const ComposePost = ({ placeholder }: ComposePostProps) => {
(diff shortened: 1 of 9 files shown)


src/components/Posts/ComposePost.tsx
```diff
@@ -35,6 +35,17 @@ const ComposePost = ({ placeholder }: ComposePostProps) => {
   const [postTextContent, setPostTextContent] = useState("");
   const user = useAppSelector((state) => state.user);
   const dispatch = useAppDispatch();
+  const buttonStyle = {
+    postButton: {
+      "&.Mui-disabled": {
```
Dennull: Can we look at putting this in our theme so that all of our disabled buttons look the same? Same thing with the border radius. Might be a good idea for us to all hop on call go through this together sometime

src/components/Posts/ComposePost.tsx
```diff
@@ -77,17 +88,17 @@ const ComposePost = ({ placeholder }: ComposePostProps) => {
           />
           <Box sx={styles.postActions}>
             <Stack direction="row">
-              <IconButton size="small">
+              <IconButton size="small" color="primary">
```
Dennull: What was the rationale for making the action buttons in these compose post/reply components always green? Should we have all of our buttons look and act the same instead?

src/components/Posts/PostItem.tsx
```diff
@@ -24,22 +24,33 @@ import { Post, updatePost } from "../../state/slices/postsSlice";
...
-};
-
 type PostProps = {
   post: Post;
 };
 
 const PostItem = ({ post }: PostProps) => {
   const dispatch = useAppDispatch();
   const user = useAppSelector((state) => state.user);
+  const styles = {
```
Dennull: I kinda don't like moving the styles inside the component like this.. I know you did it for the ternary statement, but if possible I wanna find a way to add this conditional styling while keeping the current pattern of separating our `styles` object from the component

Dennull (changes requested):
I'm a fan of the look, I think this is a good change. I left some nitpicky comments, but I think it's just kinda like that when it comes to these styling things. I'd like for us to get on call sometime to decide on these changes together

src/components/Posts/ComposePost.tsx
NasihNazeem: Sounds g lets bring this up next meeting

src/components/Posts/ComposePost.tsx
NasihNazeem: I was basically cloning twitter because they don't have it gray at any point.

It doesn't really matter to me if we make it look and act the same. Both make sense to me, so if you prefer one way over the other we can make that adjustment

src/components/Posts/PostItem.tsx
NasihNazeem: I think I figured out a neat way to get around this! Check out my next commit

src/components/Posts/ComposePost.tsx
Dennull: Did this together

src/components/Posts/ComposePost.tsx
Dennull: We ended up making all of our icon buttons follow the same pattern

src/components/Posts/PostItem.tsx
Dennull: We found a workaround for now, if anyone finds something better we'll revisit this later

Dennull (approved):

PR #55, opened 2023-06-21
T50: Fixed the PostButtonModal and multiline posts

### LINKED ISSUE #50: Styling cleanup for PostButtonModal
The `PostButtonModal` could use some styling touch ups:

- We'd like the modal to be near the top of the page rather than in the middle
- Round the corners of the modal more

There's also a bug that has to do with the multiline setting on MUI's textfield:

STEPS:

1. Click on the "post" button on the navbar
2. Enter multiple lines inside the textfield (ie. press enter multiple times)

OBSERVED:

- The modal's padding expands both above and below

EXPECTED:

- The modal's padding should only expand below, the same direction as the direction the textfield is expanding

christopherkinyua:
I fixed the Dialog positioning issue and managed to get it to be at the top of the page. I also managed to fix the expansion issue when we pressed enter multiple times to only expand downwards. This was achieved when I fixed the dialog to the top. 

I also addressed an issue that after you create a Post from not retaining the white spaces.

### DIFF
--- a/src/components/NavBar/PostButtonModal.tsx
+++ b/src/components/NavBar/PostButtonModal.tsx
@@ -4,6 +4,8 @@ import CloseIcon from "@mui/icons-material/Close";
 const styles = {
   dialog: {
     maxHeight: "90%",
+    scroll: "paper",
+    ".MuiDialog-scrollPaper": { display: "flex", alignItems: "flex-start" },
   },
   dialogContent: {
     padding: 0,
@@ -34,6 +36,7 @@ export default function PostButtonModal({
       open={openModal}
       scroll="paper"
       sx={styles.dialog}
+      PaperProps={{ sx: { borderRadius: 5 } }}
     >
       <DialogTitle sx={styles.dialogTitle}>
         <IconButton disableRipple onClick={onClose}>
--- a/src/components/NavBar/PostButtonModal.tsx
+++ b/src/components/NavBar/PostButtonModal.tsx
@@ -4,8 +4,7 @@ import CloseIcon from "@mui/icons-material/Close";
 const styles = {
   dialog: {
     maxHeight: "90%",
-    scroll: "paper",
-    ".MuiDialog-scrollPaper": { display: "flex", alignItems: "flex-start" },
+    ".MuiDialog-scrollPaper": { alignItems: "flex-start" },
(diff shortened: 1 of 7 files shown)


src/components/NavBar/PostButtonModal.tsx
```diff
@@ -4,6 +4,8 @@ import CloseIcon from "@mui/icons-material/Close";
 const styles = {
   dialog: {
     maxHeight: "90%",
+    scroll: "paper",
```
Dennull: Do we need this?

src/components/NavBar/PostButtonModal.tsx
```diff
@@ -4,6 +4,8 @@ import CloseIcon from "@mui/icons-material/Close";
...
+    ".MuiDialog-scrollPaper": { display: "flex", alignItems: "flex-start" },
```
Dennull: Do you need `display: "flex"`? I think that is already applied as default from the MUI component

src/components/NavBar/PostButtonModal.tsx
```diff
@@ -34,6 +36,7 @@ export default function PostButtonModal({
       open={openModal}
       scroll="paper"
       sx={styles.dialog}
+      PaperProps={{ sx: { borderRadius: 5 } }}
```
Dennull: Can we look into doing this in `Theme.tsx` so it applies to all of our dialogs?

Dennull (changes requested):
A few things to clean up, but I like the changes! I also wanna change some padding values, so if we can do that over a call that'd be great

src/components/NavBar/PostButtonModal.tsx
christopherkinyua: Nope, you are right. Took it off

src/components/NavBar/PostButtonModal.tsx
christopherkinyua: Yes, it is the default. I removed it.

src/components/NavBar/PostButtonModal.tsx
christopherkinyua: Done

Dennull (approved):

PR #59, opened 2023-07-07
T46: Convert timestamps

### LINKED ISSUE #46: Convert timestamps into human readable date
The timestamps in our app need to be converted to human readable dates. We should look into if we can leverage Javascript date libraries (ex. date-fns) to do this work for us

AnindaS13:
This PR is for the timestamps to be converted to appropriate times. When a post is just posted, it says "just now". Otherwise it uses minutes if it was posted less than an hour ago. It posts hours if it was more than 1 hour. And then the specific date if it is older than a day. Also the post padding was moved a bit. Let me know if it looks fine.


src/components/Profile/ProfileLikes.tsx
```diff
@@ -19,7 +20,11 @@ const ProfileLikes = () => {
           },
         }
       );
-      dispatch(setPosts(result.data as Post[]));
+      const modified = result.data.map((entry: any) => ({
```
NasihNazeem: This works well, but should we start separating the fetch from logical changes? Like throwing the fetch in its own method being called in the `useEffect` rather than just having the fetch in the `useEffect`?

src/components/Profile/ProfileLikes.tsx
```diff
@@ -19,7 +20,11 @@ const ProfileLikes = () => {
...
+        ...entry,
+        timestamp: formatTimestamp(entry),
+      }));
+      dispatch(setPosts(modified as Post[]));
     };
     fetchPosts();
   }, [dispatch, user]);
```
NasihNazeem: Because `formatTimestamp` exists outside the scope of this `useEffect` we should add it to the dependency array
```suggestion
  }, [dispatch, user, formatTimestamp]);
```

src/components/Posts/ExpandedPostList.tsx
NasihNazeem, line 1: I don't think we need this page. This seems to be an unused and outdated version of `ExpandedPostReplies.tsx`

src/components/Posts/ExpandedPostReplies.tsx
```diff
@@ -25,13 +26,17 @@ const ExpandedPostReplies = ({ post }: ExpandedPostRepliesProps) => {
...
+          timestamp: formatTimestamp(entry),
+        }));
+        dispatch(setPosts(modified as Post[]));
       } catch (e) {
         console.log(e.message);
       }
     };
     fetchPosts();
-  }, [dispatch, user, post]);
+  }, [dispatch, user, post, posts]);
```
NasihNazeem: I think we should revert this back to how it was originally because adding `posts` to the dependency array seems to infinitely call the `useEffect` for some reason. Try `console.log` in the useEffect and tell me if you get the same result

src/components/Profile/ProfileLikes.tsx
NasihNazeem: And it doesn't seem to have any effect on the logic inside the useEffect

src/components/Posts/ExpandedPostReplies.tsx
```diff
@@ -25,13 +26,17 @@ const ExpandedPostReplies = ({ post }: ExpandedPostRepliesProps) => {
             },
           }
         );
-        dispatch(setPosts(resultReplies.data as Post[]));
+        const modified = resultReplies.data.map((entry: any) => ({
```
NasihNazeem: Instead of using `any` as the type for `entry` we should use `Post`. Using `any` feels like it takes away from using Typescript and generally isn't really used

NasihNazeem:
```suggestion
        const modified = resultReplies.data.map((entry: Post) => ({
```

src/components/NavBar/formatTimestamp.tsx
```diff
@@ -0,0 +1,28 @@
+const formatTimestamp = (entry: any) => {
```
NasihNazeem: Let's not use `any` here, instead let's use `Post`. Using `any` takes away from Typescript

```suggestion
const formatTimestamp = (entry: Post) => {
```

src/components/NavBar/formatTimestamp.tsx
NasihNazeem, line 1: Overall this is good! I noticed that this implementation can be further simplified using a library called `moment.js`. It handles a lot of the duration calculations and we should try to leverage more libraries if they can help us

NasihNazeem (changes requested):
Have some changes to make for this PR but overall great job. If you are busy because you're out of the country I don't mind taking on this branch to finish up the PR as well.

I wanted to suggest talking about a systemic change as well because having to change EVERY fetch to `modified` must have been annoying and I feel like we should talk about whether that is still something we want to do because it feels like there is a better way to do that

src/components/NavBar/formatTimestamp.tsx
Dennull: I agree, we should look into if we can use libraries to simplify these date calculations for us. We can merge this once it's ready and make another ticket for that.

One other thing is that since this isn't a component, the file extension should be `.ts` rather than `.tsx`. We should also make a utilities folder and put this file in there for the same reason

src/components/NavBar/formatTimestamp.tsx
Dennull: Nice suggestion. I also don't think the type should be a `Post` either. Posts aren't the only place where we display timestamps, so I think we should pass in the timestamp itself rather then the whole post. In this case the type would be a `string`

Dennull (changes requested):
Nice work on the function! However, I think that the way we've implemented the function here could be improved on. Rather than using `formatTimestamp` before every dispatch, we should leave timestamps as they are in the store and only change them upon render in the components

For example, for expanded posts, instead of changing the timestamp in the dispatch function, we should call the function in `ExpandedPostItem.tsx` when we render it. If that's confusing, I can hop on call to explain what I mean better

src/components/Profile/ProfileLikes.tsx
AnindaS13: I think this is something we do in all our components, so maybe we could discuss it in our meeting and make a different ticket for it.

src/components/NavBar/formatTimestamp.tsx
Dennull: As discussed, we'll merge these changes in for now and make a new ticket to explore if we want to use a third party library

Dennull (approved):
LGTM! I made a few styling changes, so @NasihNazeem and @AnindaS13 if you guys are good with them feel free to merge

NasihNazeem (approved):

PR #61, opened 2023-07-18
T58: Page layout template

### LINKED ISSUE #58: Create consistent page layout
Since the majority of our pages have the same layout of navbar on the left, main content in the middle, and secondary content on the right, it'd be great to have a component that keeps this layout consistent for us. Bonus points if we can set the width of each section using a prop since a small number of our pages may differ in terms of the width of each section (ex. messages page)

Dennull:
Resolves #58. This PR introduces a new `Layout` component that we should wrap our pages with to keep their layouts consistent. The component accepts two props:
- `middleContent` is usually the main content that's displayed in the middle of the page
- `rightContent` is usually secondary content that's displayed on the right side of the page

`Layout` takes care of layout styling by keeping the navbar, `middleContent` and `rightContent` the same size in all situations

The only pages that don't use `Layout` are `Messages` and `DirectMessages`. Custom layouts were made for the messages pages following the same principles as the others; the only difference is that the middle content is smaller than the right content in these pages

### DIFF
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -50,6 +50,7 @@ const styles = {
     paddingLeft: 0.5,
     fontSize: 14.5,
   },
+  backButton: { "&:hover": { backgroundColor: "transparent" } },
   card: {
     padding: 0,
     boxShadow: "none",
@@ -143,7 +144,7 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
   return (
     <Card sx={styles.card}>
       <Box style={styles.topHeader}>
-        <IconButton onClick={() => navigate(-1)}>
+        <IconButton onClick={() => navigate(-1)} sx={styles.backButton}>
           <KeyboardBackspaceIcon color="secondary" />
         </IconButton>
         <Typography style={styles.headerTitle}>Post</Typography>
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -48,10 +48,6 @@ const styles = {
     paddingLeft: 0.5,
     fontSize: 14.5,
   },
-  backButton: {
-    backgroundColor: "transparent",
-    paddingRight: 10,
-  },
   card: {
     padding: 0,
     boxShadow: "none",
@@ -141,7 +137,7 @@ const ExpandedPostItem = () => {
   return (
     <Card sx={styles.card}>
       <Box style={styles.topHeader}>
-        <IconButton style={styles.backButton} onClick={() => navigate(-1)}>
+        <IconButton onClick={() => navigate(-1)}>
           <KeyboardBackspaceIcon color="secondary" />
         </IconButton>
         <Typography style={styles.headerTitle}>Post</Typography>
--- a/src/pages/DirectMessage.tsx
+++ b/src/pages/DirectMessage.tsx
@@ -251,7 +251,6 @@ const DirectMessage = () => {
           </Box>
         </Box>
       </Box>
-      <Divider orientation="vertical" />
     </Stack>
   );
 };
--- a/src/pages/DirectMessage.tsx
+++ b/src/pages/DirectMessage.tsx
@@ -32,6 +32,7 @@ const styles = {
   container: { height: "auto", justifyContent: "center" },
   chatContainer: {
     display: "flex",
+    flex: 1,
     flexDirection: "column",
     overflowY: "hidden",
   },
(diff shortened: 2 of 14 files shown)


src/pages/Layout.tsx
NasihNazeem, line 1: I'm a fan of this change. It abstracts the "boilerplate" code from other pages and obv fixes the layout of each page

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -147,7 +143,7 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
   return (
     <Card sx={styles.card}>
       <Box style={styles.topHeader}>
-        <IconButton style={styles.backButton} onClick={() => navigate(-1)}>
+        <IconButton onClick={() => navigate(-1)}>
```
NasihNazeem: Can we add a `disableRipple` to this button? Think it looks better when hovering

This would apply to other pages as well

src/pages/DirectMessage.tsx
```diff
@@ -148,95 +152,103 @@ const DirectMessage = () => {
...
-                }}
-                onChange={(e) => setTextContent(e.target.value)}
-                placeholder="Send a message"
-                size="small"
-                value={textContent}
-              />
-            </Box>
-          </form>
+          <Box sx={styles.chatContainer}>
+            <List component="div" ref={messageRef} sx={styles.messageList}>
```
NasihNazeem, line 203: There is a bug with the size of this component, as the height is pinned to be a minimum depending on how many messages there are.

src/pages/DirectMessage.tsx
```diff
@@ -148,95 +152,103 @@ const DirectMessage = () => {
...
+                  onChange={(e) => setTextContent(e.target.value)}
+                  placeholder="Send a message"
+                  size="small"
+                  value={textContent}
+                />
+              </Box>
+            </form>
+          </Box>
         </Box>
       </Box>
```
NasihNazeem, line 274: Can we add a divider here so that the Chat component has another outline on the RHS

src/pages/Register.tsx
NasihNazeem, line 1: I'd like to go through this page with another ticket to explore other designs, and maybe fill up the screen a bit more.

Low priority

src/pages/Messages.tsx
```diff
@@ -1,39 +1,53 @@
...
+            <Typography variant="h6">Select a Message</Typography>
+            <Typography>
+              Choose one of your existing conversations or start a new one!
+            </Typography>
+            <Button variant="contained" sx={styles.button}>
+              New Message
+            </Button>
+          </Box>
         </Box>
       </Box>
```
NasihNazeem: Similar to my comment on the `DirectMessages` page, I think adding a `Divider` on the RHS (of the Chat container) would be beneficial. It makes the app look more centered, rather than one component of the page being extremely larger than the rest

src/pages/DirectMessage.tsx
NasihNazeem: I found that using `flex: 1` on both the `chatContainer` and `messageList` styles fixed this bug.

NasihNazeem (changes requested):
Nice work! I'm sure this was a pain to work through because of so many unpolished changes with every new feature. I think this sets us up for easier integration of other features and mitigates the # of bugs we might face with older features.

I left some comments where I found some more bugs to do with the new changes, feel free to take a look. I would like to review this again anyways because its a large change. I have some more changes I would like to suggest that don't have to do with one page in specific: 

- There was one global bug that has to do with modals and the global scrollbar. When you open up a modal, the scrollbar disappears. Other than that I found no real bugs with that scrollbar.

- We should have another ticket more focused on the responsiveness of the page later on when the app is more complete. It would be better to do it later because we can adjust for any upcoming features along with the old stuff instead of focusing on that with every new feature.

Let me know your thoughts!

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: Not `disableRipple`, but remove the hover background

Dennull: Sounds good

src/pages/DirectMessage.tsx
Dennull: Nice catch!

src/pages/DirectMessage.tsx
Dennull: As discussed offline, there's some complication around this since it adds an extra pixel, which makes the layout of the messages pages 1 pixel off compared to the other pages. Hopefully we can solve this in another ticket

src/pages/Messages.tsx
Dennull: As discussed offline, there's some complication around this since it adds an extra pixel, which makes the layout of the messages pages 1 pixel off compared to the other pages. Hopefully we can solve this in another ticket

Dennull: Same here as in `DirectMessage.tsx`

NasihNazeem (approved):

PR #64, opened 2023-07-21
T56: Fix expanded post actions styling

### LINKED ISSUE #56: Expanded post actions shift when the numbers change
STEPS:

1. Click on an expanded post
2. Like a post

OBSERVED:
- Since the like count changes, the text of the nearby buttons (ex. "likes", "replies", etc) also shift, which looks kind of ugly
- The spacing between the like action button and the reply action button is inconsistent with the rest of the spacing

EXPECTED:
- The individual text beside the numbers shouldn't shift. It would look better if the whole button shifts rather than the text
- Fix the spacing between the like action button and the reply action button. The spacing between all the buttons should be the same

Dennull:
Resolves #56. These changes were made to clean up the look of the expanded post actions

Before, when the numbers beside the action buttons changed, the individual text (ex. "likes") beside the number would all shift. This has been changed so that the whole button shifts rather than the individual text

The action buttons and corresponding icons were also reordered. Buttons are now ordered from most likely to be used to less likely to be used to minimize shifting

### DIFF
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -211,6 +211,16 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
           justifyContent="space-between"
           sx={styles.cardActions}
         >
+          <IconButton>
+            <RepeatOutlinedIcon />
+          </IconButton>
+          <IconButton
+            onClick={() => {
+              setOpen(true);
+            }}
+          >
+            <AddCommentOutlinedIcon />
+          </IconButton>
           <IconButton
             onClick={() => {
               post.isLikedByCurrentUser
@@ -224,16 +234,6 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
             ) : (
               <FavoriteBorderOutlinedIcon />
             )}
-          </IconButton>
-          <IconButton
-            onClick={() => {
-              setOpen(true);
-            }}
-          >
-            <AddCommentOutlinedIcon />
-          </IconButton>
-          <IconButton>
-            <RepeatOutlinedIcon />
           </IconButton>
           <IconButton>
             <ShareOutlinedIcon />
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -32,23 +32,16 @@ import { Post } from "../../state/slices/postsSlice";
 
 const styles = {
   actionButton: {
-    fontSize: 14.5,
-    padding: 0,
     color: "black",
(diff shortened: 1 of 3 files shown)


src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -164,28 +158,48 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
...
-        </Button>
-        <Button component={"span"} sx={styles.actionButton}>
-          1<Typography sx={styles.actionTitles}>Comments</Typography>
-        </Button>
-        <Button component={"span"} sx={styles.actionButton}>
-          1<Typography sx={styles.actionTitles}>Reposts</Typography>
-        </Button>
+        <Box>
+          <Button fullWidth sx={styles.actionButton}>
+            <Typography component="span" sx={{ fontWeight: "bold" }}>
```
NasihNazeem: Lets put the `sx` in the `styles` object. This goes for all three uses

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -59,21 +52,22 @@ const styles = {
...
-    paddingLeft: 2,
     fontSize: 14.5,
     color: "#a4a8ab",
   },
   timestampBox: {
     display: "flex",
-    paddingTop: 1,
     paddingBottom: 1,
+    paddingLeft: 2,
+    paddingTop: 1,
```
NasihNazeem: `paddingY: 1` works here

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -32,23 +32,16 @@ import { Post } from "../../state/slices/postsSlice";
 
 const styles = {
   actionButton: {
```
NasihNazeem, line 34: We should remove the hover background from the action buttons
```suggestion
 actionButton: {
     "&:hover": {
         backgroundColor: "transparent",
      }
```

NasihNazeem (changes requested):
Thanks for doing this, should have been fixed earlier but this will stop bugging us from here on out

src/components/Posts/ExpandedPostItem.tsx
Dennull: Yeah, good one!

src/components/Posts/PostList.tsx
```diff
@@ -26,13 +27,13 @@ const PostList = () => {
   }, [dispatch, user]);
 
   return (
-    <>
+    <Stack divider={<Divider />}>
```
NasihNazeem, line 30: Nice change

NasihNazeem (approved):
Sweet, merge with when you're free

PR #66, opened 2023-07-24
T38: Right Side Content

### LINKED ISSUE #38: Right side content
The right side content of most of our pages will look the same. For MVP, this should have:

- An ad space with a dummy ad
- Follower suggestions
- A footer that can container dummy links
- A dummy search bar

NasihNazeem:
This PR resolves #38. This introduced the Sidebar on `Timeline`, `Profile` and `ExpandedPost`. 

It includes features such as:
- General Search Bar
- Suggested Users
- Relevant People (exclusive to ExpandedPost)
- Advertisements
- Footer Pages

The sidebar for `Timeline` and `Profile` is different from the one displayed within `ExpandedPost` by removing the SuggestedUsers component and introducing a Relevant People tab where the user can quickly follow the author of the post.

This PR also introduces a `Coming Soon` page that we can temporarily use for all Link and Button redirects to pages and features that have not been implemented yet. This has been implemented in most components within the Sidebar. If the feature is something we cannot implement for whatever reason, I suggest using this or something like this.

There has been an update to `Theme.tsx` where I have added a light color of `gray`. We didn't discuss this officially, so if we want to change the hex code, we can discuss this.

Lastly, it's important to state that this PR is just a foundation for the features being implemented within the sidebar. Many of the features have not been implemented yet due to out-of-scope constraints and will require new issues for implementation.


public/Ad Gray.gif
Dennull, line 1: Lmao I love this

src/components/SideBar/ExpandedPostSidebar.tsx
```diff
@@ -0,0 +1,243 @@
...
+        <Box sx={styles.ruHeader}>
+          <Typography variant="h6" sx={styles.ruTitle}>
+            Relevant People
+          </Typography>
+        </Box>
+        <Box sx={styles.ruContent} className="ruContent">
+          <Box sx={styles.userInfo}>
+            <Avatar />
+            <Box sx={styles.ruNames}>
+              <Link
```
Dennull: Instead of using `href`, try passing a react-router-dom `Link` to the `component` prop and adding the route to the `to` prop. You can look at some of our other usages of links or https://mui.com/material-ui/guides/routing/ for examples. This applies to other usages of links for redirects too

src/components/SideBar/SideBar.tsx
```diff
@@ -0,0 +1,179 @@
...
+};
+
+const SideBar = () => {
+  return (
+    <Box sx={styles.rightContent}>
+      <Box sx={styles.searchBarContainer}>
+        <SearchBar placeholder="Search Chirp" />
+      </Box>
+
+      <Box sx={styles.suggestedUserContainer}>
```
Dennull: I think we can put this header into `SuggestedUsers`

src/components/SideBar/ExpandedPostSidebar.tsx
Dennull, line 1: There seems to be a lot of unused styles here. Try to double check to see if the things in the `styles` object actually are needed. I see some used of `className` in some places too, but we don't use css so let's get rid of those

I'd also like to abstract more things in this page. I think the relevant users list and the footer can be their own components

src/components/SideBar/SideBar.tsx
Dennull, line 1: Similar comments to `ExpandedPostSidebar.tsx`. Please double check for unused styles and delete uses of `className`. I think the footer can be it's own component too

Try to avoid using `href` here and in the `ExpandedPostSidebar` since it's better to use react router's stuff

src/components/SideBar/SuggestedUsers.tsx
```diff
@@ -0,0 +1,121 @@
...
+  },
+];
+
+const SuggestedUsers = () => {
+  const navigate = useNavigate();
+
+  return (
+    <List sx={styles.list}>
+      {usersData.map((user) => (
+        <ListItem key={user.id} sx={styles.userContainer}>
```
Dennull: Since this isn't a button, I'd want to remove the pointer styling on the cursor since it's misleading. We could turn this into a `ListItemButton` so that users can click anywhere on the button too, but I think MUI advises against having buttons within buttons because of overlapping `onClick` functions.

Personally, I'd like it if we could make the whole thing a button but not sure if we should go against what MUI recommends. Maybe try it out and see if you run into any issues?

src/pages/Layout.tsx
```diff
@@ -11,7 +11,6 @@ const styles = {
     boxSizing: "border-box",
     flex: "0 0 350px",
     minWidth: 0,
-    paddingLeft: "40px",
```
Dennull: Can we bring this back? It saves us from having to add padding individually to whatever the right content is

Dennull (changes requested):
Love the way this looks, it makes the app feel more complete. I have a few suggestions about creating components for some things and a few comments on how routing is done. Also could use some cleanup for unused styling. Nice job though!

src/components/SideBar/SuggestedUsers.tsx
NasihNazeem: I applied `ListItemButton` and for some reason this doesn't give me the warning where we have a `Button` as a descendant of a `Button`. So I will permanently change it to `ListItemButton`. 

But even though there was no warning, there is still the issue of overlapping `onClick` functionality so I had to include an `onClick` handler for the child button (follow button) that stops the clicking propagation at the child and doesn't go through the DOM to the ancestor components.

src/components/SideBar/ExpandedPostSidebar.tsx
NasihNazeem: Applied this and removed any unnecessary `Link` calls as well

src/components/SideBar/SideBar.tsx
NasihNazeem: Moved reusable containers such as `SuggestedUsers` into components. This includes the `SidebarFooter` and `RelevantUsers`.

src/components/SideBar/ExpandedPostSidebar.tsx
NasihNazeem: Removed all `classNames` within the app, including out-of-scope components. Removed unused styles as well.

src/components/SideBar/SuggestedUsers.tsx
Dennull: Nice! I like it

Dennull (approved):

PR #69, opened 2023-07-26
T68: Display reply and repost count

### LINKED ISSUE #68: Display actual count of replies
Now that users can reply to posts, we should update the reply count beside the reply icon on our post lists and expanded posts to display the actual number of replies rather than a hard coded value

Dennull:
This resolves #68. The actual number of replies and reposts are now displayed on posts and expanded posts

Some refactors were also done to simplify our post state management on the front end:
- `postsSlice.ts` was refactored to move all of our front end post state logic into the reducers
- `expandedPostSlice.ts` was removed. The state and state logic were moved into `postsSlice.ts`
- The POST and DELETE requests for liking and deleting a post respectively were moved into a utility file called `postUtilities.ts`


src/pages/Profile.tsx
NasihNazeem, line 1: Good styling updates

src/state/slices/postsSlice.ts
NasihNazeem, line 1: Good work abstracting the functionality of the dispatch away from the components and into reducers

src/state/slices/postsSlice.ts
```diff
@@ -14,31 +16,98 @@ export type Post = {
...
   reducers: {
-    setPosts: (_, action: PayloadAction<Post[]>) => {
-      return action.payload;
-    },
-    updatePost: (state, action: PayloadAction<Post>) => {
-      const newPosts = state.map((o) => {
-        if (o.postId === action.payload.postId) {
-          return action.payload;
+    addReply: (state, action: PayloadAction<Post>) => {
+      const newReply = {
```
NasihNazeem, line 46: Nice work, since you've included `numberofReposts` maybe we should add this to the `addReply` and `appendPost` reducers similar to what you've done with `numberofReplies`.

Though it is considered out of scope and could be done in another issue that has to do with the reposting functionality
```suggestion
      const newReply = {
          ...action.payload,
          numberOfLikes: 0,
          numberOfReplies: 0,
          numberOfReposts: 0,
      };
```

src/utilities/postUtilities.ts
NasihNazeem, line 1: Nice separation for this function. For future functions, would something like reposting toggling functionality be created in this utility folder as well?

I assume reply functionality can't because it's a form and the abstraction might be more complicated than what we already have

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -177,31 +141,25 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
...
         <Box>
           <Button fullWidth sx={styles.actionButton}>
             <Typography component="span" sx={styles.actionCount}>
-              1
-            </Typography>
-            <Typography component="span" sx={styles.actionTitles}>
-              Replies
+              {post.numberOfReplies}
             </Typography>
+            <Typography component="span">Replies</Typography>
```
NasihNazeem: Can we reduce the font size of these action buttons? If it was the same font size as the timestamp right above it would look better

src/utilities/postUtilities.ts
NasihNazeem: Further question, why not keep it as a method within the same component?

src/state/slices/postsSlice.ts
```diff
@@ -14,31 +16,98 @@ export type Post = {
...
+            numberOfLikes: isLikedByCurrentUser
+              ? o.numberOfLikes + 1
+              : o.numberOfLikes - 1,
+          };
+        }
+        return o;
+      });
+      state.posts = newPosts;
+
+      if (action.payload === state.expandedPost.postId) {
```
NasihNazeem, line 97: Nice, this removes the need for `updateExpandedPost` reducer I think

NasihNazeem (changes requested):
Good work! I liked seeing the merge for expandedPostSlice because that always felt forced and unnecessary. I left some comments with regard to to some minor changes, and some questions to understand the reasoning behind some of the programming choices.

The functionality works as expected as well so good job

src/state/slices/postsSlice.ts
Dennull: Yea I forgot about this, I'll add it in

src/utilities/postUtilities.ts
Dennull: I put the function in this file because it's used in 2 components: `PostItem.tsx` and `ExpandedPostItem.tsx`

src/components/Posts/ExpandedPostItem.tsx
Dennull: Sure! I don't like how we'll have to apply a style to each span, but that's part of a bigger issue with our overall typography, so let's worry about cleaning that up later

src/state/slices/postsSlice.ts
Dennull: Yup, `updateExpandedPost` was part of `expandedPostSlice.ts`, which got deleted

NasihNazeem (approved):

PR #72, opened 2023-07-30
T38: Styling feedback for right side content

Dennull:


src/components/SideBar/SidebarFooter.tsx
```diff
@@ -1,88 +1,75 @@
...
-    },
+  footerText: {
     color: "gray.main",
     fontSize: 12,
+    paddingRight: 1.5,
   },
   footerNav: {
     display: "flex",
     flexWrap: "wrap",
-    paddingX: 2,
```
NasihNazeem, line 20: I prefer to keep this aligned to the left, centering it looks awkward with the other text in the sidebar being aligned left

src/components/SideBar/FollowButton.tsx
```diff
@@ -1,29 +1,32 @@
...
-      backgroundColor: "#1E9964", // could be primary.dark
+      backgroundColor: "#1E9964", // Could be primary.dark
     },
   },
 };
 
-const FollowButton = () => {
-  const handleFollowButton = (e: React.MouseEvent) => {
+type FollowButtonProps = {
+  onClick?: () => void;
```
NasihNazeem, line 15: I am confused with how we want to use this onClick

I was assuming the state of the follow button would be handled within this component

src/components/SideBar/FollowButton.tsx
```diff
@@ -1,29 +1,32 @@
...
-import React from "react";
 
 const styles = {
   followButton: {
     boxShadow: "none",
     backgroundColor: "primary.main",
     "&:hover": {
       boxShadow: "none",
-      backgroundColor: "#1E9964", // could be primary.dark
+      backgroundColor: "#1E9964", // Could be primary.dark
```
NasihNazeem: I will make this color `primary.dark` if there are no objections

src/components/SideBar/SideBar.tsx
```diff
@@ -20,7 +19,7 @@ const styles = {
 const SideBar = () => {
   return (
     <Box sx={styles.rightContent}>
-      <SidebarSearchBar />
+      <SearchBar placeholder="Search Chirp" />
```
NasihNazeem: Instead of having a `paddingTop` for the entire Sidebar, let's do something similar to what we did with the search bar `ConversationList.tsx` where we provided a container for the search bar and aligned it using that.

I prefer that method rather than padding the entire content container. However, you did say you rather have it without the `Box`, but this is up to you

src/components/SideBar/SidebarFooter.tsx
```diff
@@ -1,88 +1,75 @@
...
+    justifyContent: "center",
   },
 };
 
 const SidebarFooter = () => {
   return (
-    <Box component={"nav"} sx={styles.footerNav}>
+    <Box component="nav" sx={styles.footerNav}>
       <Link
```
NasihNazeem, line 19: I rather have links in the footer open a new tab rather than open in the same tab because of the UX in our app. 

We should not force users to lose their place on our website, because they could be scrolling far and navigating back might make it more difficult to find that place.

NasihNazeem (changes requested):
Good changes in this PR, I had some comments on some the changes

src/components/SideBar/FollowButton.tsx
Dennull: I made this prop as an optional one in case you wanted to add an onClick handler to the button. Sometimes you also want other actions done when someone clicks the button so this can be used for that

src/components/SideBar/SideBar.tsx
Dennull: Sure, no objection here

src/components/SideBar/SidebarFooter.tsx
Dennull: Agreed

NasihNazeem (approved):

PR #76, opened 2023-08-16
T79: Update profile routes, change default profile picture and banner

### LINKED ISSUE #75: Provide placeholder avatar and banner
Change the placeholder profile picture to be a grey avatar and change the profile banner to be a grey background. These will be the default profile picture and banner for new accounts

NasihNazeem:
This PR resolves #75. This updates the profile page route to a variable that is a selected user's `username`. There has also been an update to the profile picture and banner to be more of a template rather than customizable. Lastly, we have updated the requests to send the username which as been passed as a parameter in the URL. This also introduces changes to provide routes from each `displayName` found on the app and any avatar

Lastly, stylistic updates to the profile page include a border to the profile picture and divider updates to the `ExpandedPostReplies` and `PostItem`


src/components/Common/FollowButton.tsx
```diff
@@ -23,8 +26,8 @@ const FollowButton = ({ onClick }: FollowButtonProps) => {
 
   return (
     <Button
-      onClick={handleFollow}
```
Dennull: Nitpick, can we move this back to keep it alphabetical?

src/components/Posts/ComposePost.tsx
```diff
@@ -66,7 +74,9 @@ const ComposePost = ({ placeholder, minRows }: ComposePostProps) => {
     <form onSubmit={onSubmit}>
       <Box sx={styles.compostPostContainer}>
         <Box sx={styles.avatarIcon}>
-          <Avatar />
+          <Link component={Routerlink} to={`/${user.username}`}>
+            <Avatar sx={styles.avatar} />
+          </Link>
```
Dennull: This avatar link styling and markup is repeated a lot, I think it'd be good to make a component for this

src/components/Posts/ExpandedPostReplies.tsx
```diff
@@ -34,13 +35,14 @@ const ExpandedPostReplies = ({ post }: ExpandedPostRepliesProps) => {
   }, [dispatch, user, post]);
 
   return (
-    <>
+    <Stack divider={<Divider />}>
```
Dennull: Small thing but not sure if we should use a `Stack` here.. I remember we tried to stray away from using the `Stack` component for consistency's sake unless there was a merit to using it

We do want a `Divider` here but since we also want a `Divider` on the last element too I wonder if we can get away with adding a `Divider` within the map on line 43 after each `PostItem`, or modifying the border of the `PostItem`

Not worth blocking the PR over but maybe we can hop on a call sometime and experiment?

src/components/Profile/ProfileLikes.tsx
```diff
@@ -3,33 +3,36 @@ import PostItem from "../Posts/PostItem";
...
 
   useEffect(() => {
     const fetchPosts = async () => {
       const result = await axios.get(
-        "http://localhost:3001/api/profile/getOwnLikes",
+        "http://localhost:3001/api/profile/getUserLikes",
         {
           params: {
-            userId: user.userId,
+            username,
```
Dennull, line 20: This makes me wonder if we should be using userId's or usernames for querying user info in the backend. It seems weird to use them interchangeably. I feel like we should be picking one and sticking with it for consistency

Now that I think about it I don't know why we have both a userId and a username, I think I just copied what I saw other people do. There must be a reason it's like that tho, maybe it's a performance thing? We should look into it and find out if it's better to query by username or by userId

I do see why we'd want the profile route to be /:username tho cause it makes it easier for people to type in a profile they're interested in. But maybe what ends up happening is we make the route /:username but query our API using the userId

src/pages/Profile.tsx
```diff
@@ -98,7 +138,8 @@ const Profile = () => {
       });
     };
     fetchProfileContents();
-  }, [value, user]);
+    window.scrollTo(0, 0);
```
Dennull, line 137: Does this scroll to the top? Is it because when you click on a new profile you want it to always be on the top? At the same time if you were just viewing someone else's profile and press the back button you might want to be where you where, so that might complicate things

src/pages/Profile.tsx
```diff
@@ -109,42 +150,38 @@ const Profile = () => {
               <KeyboardBackspaceIcon color="secondary" />
             </IconButton>
             <Box>
-              <Typography sx={styles.displayName}>
-                {user.displayName}
+              <Typography sx={styles.headerName}>
```
Dennull: Do you think we should keep the look of the display name in the header the same as it is in the bio area?

src/pages/Profile.tsx
```diff
@@ -1,38 +1,62 @@
...
     alignItems: "flex-start",
     display: "flex",
     justifyContent: "space-between",
   },
-  banner: { width: "100%", height: "200px" },
+  banner: {
+    width: "100%",
+    height: "200px",
+    backgroundColor: "primary.main",
+    borderSizing: "border-box",
```
Dennull: Do we need this box-sizing property?

src/pages/Profile.tsx
```diff
@@ -1,38 +1,62 @@
...
+import { useNavigate, useParams } from "react-router-dom";
 import IconButton from "@mui/material/IconButton/IconButton";
 import Layout from "./Layout";
 import SideBar from "../components/SideBar/SideBar";
+import { Link as Routerlink } from "react-router-dom";
+import { useAppSelector } from "../state/hooks";
+import FollowButton from "../components/Common/FollowButton";
 
 const styles = {
   avatar: {
```
Dennull, line 29: I like the border change! Can we turn the height and width back to what it was before? The issue you probably ran into is that the `Avatar` component styles differently so you have to add `px` after the number. We should also make the box sizing border-box since we're adding a border:

```suggestion
  avatar: {
    border: "5px solid white",
    boxSizing: "border-box",
    height: "140px",
    marginTop: "-15%",
    width: "140px"
  },
```

Dennull (changes requested):
I had a few suggestions. I also have some small gripes about a few styling things, but those might fall into some typography stuff. More specifically they are:

- The look of the follow button and edit profile button
- The font of the username and display name

We already have another ticket for typography stuff so maybe we can leave these for then, but I'd be down to get on a call about them tho!

src/components/Posts/ComposePost.tsx
NasihNazeem: Done, and I also introduced it to other components like `ConversationListItem` where we had an avatar. Same changes to `DirectMessages`, `RelevantUsers`, `ComposePost`, `ComposeReply`

src/components/Posts/ExpandedPostReplies.tsx
NasihNazeem: Adding a `Divider` within the map caused some issues with React where I would have used a unique key on two components, which gave a warning.

I think we can experiment, so I'll remove the `Stack`, and leave the `Divider` that was already on line 44 and we can fix this up soon

src/pages/Profile.tsx
NasihNazeem: I'd like it so that it resets to the top because if you don't have that scroll to the top on the profile page, it can cause an annoying bug where:

Let's say you have scrolled down on the timeline page, and want to visit a user's profile. Now because you have already scrolled down, the scrollbar keeps its state even on the profile page and it leads to bad UX because it can get confusing unless you are redirected to the top of the profile page.

With all that being considered, I see your point and honestly can't think of a solution for both to coexist just yet

src/pages/Profile.tsx
NasihNazeem: I'll actually revert this because it'd be better to keep it the same, but idk if you see this but I feel like there's an uneven amount of space between the top and bottom borders of the header and the text

src/pages/Profile.tsx
NasihNazeem: Nope, removing this

src/pages/Profile.tsx
NasihNazeem: Implemented!

src/components/Profile/ProfileLikes.tsx
NasihNazeem: So I have updated the profile route to be using `/:username` however, I think if we want to make a decision for removing the use for `userId`, we should discuss later on and then finalize how we plan on using/not using them in another ticket

src/pages/Profile.tsx
NasihNazeem: Looking back, I think introducing a component that takes care of "scroll to top" at the App level would be better. I assume we want this in most button clicks, UNLESS it is the back button, in which case, we should can incorporate a useEffect to scroll back to a set Scroll position that we need to set beforehand.

What do you think?

PR #78, opened 2023-11-13
T77: Conversion from CRA to Vite

### LINKED ISSUE #77: Migration to Vite from CRA
This issue focuses on the migration to Vite from Create React App. Vite provides many advantages over CRA such as 

- Quicker development server
- Production builds are much more optimized due to the usage of the Rollup bundler instead of Webpack

I will be using this [article](https://www.darraghoriordan.com/2021/05/16/migrating-from-create-react-app-to-vite) to help with migration

To learn more click on the following hyperlinks: [Rollup](https://rollupjs.org/), [Vite vs Webpack](https://kinsta.com/blog/vite-vs-webpack/), [Why Vite over CRA](https://muzammilbilwani.medium.com/why-you-should-prefer-vite-over-create-react-app-cra-5adf3e71771f)

NasihNazeem:
This resolves #77. The changes in this file are as described below:

- Removal of `react-scripts` from the `package.json` file
- Removal of old boilerplate files associated with CRA such as `manifest.json` and `robots.txt`
- Introduction of Vite dependencies into `package.json`
- Creation of `vite.config.ts` and `vite-env.d.ts`
- Updated `tsconfig.ts` to update types
- Updated `index.html` to include a script module and moved it to the root folder
- Updated `index.js` to `index.tsx`
- Updated all environment variable imports


package.json
```diff
@@ -14,7 +14,6 @@
     "@testing-library/react": "^13.4.0",
     "@testing-library/user-event": "^13.5.0",
     "@types/jest": "^27.5.2",
-    "@types/node": "^16.18.11",
```
Dennull: Was this removed on purpose? Do we need this?

package.json
Dennull, line 1: Did you remove CRA? I think we can run `npm uninstall react-scripts` to do that

src/index.tsx
```diff
@@ -8,7 +8,9 @@ import AuthProvider from "./components/Auth/AuthProvider";
 import { Provider } from "react-redux";
 import { store } from "./state/store";
 
-const root = ReactDOM.createRoot(document.getElementById("root"));
+const root = ReactDOM.createRoot(
+  document.getElementById("root") as HTMLElement
```
Dennull, line 12: Out of curiosity do you know why typescript didn't complain about this before?

vite.config.ts
```diff
@@ -0,0 +1,24 @@
...
+import svgrPlugin from "vite-plugin-svgr";
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  // This changes the out put dir from dist to build
+  // comment this out if that isn't relevant for your project
+  build: {
+    outDir: "build",
+  },
+  server: {
```
Dennull, line 12: Cool thing I found, if we add `open: true` to this server object I think the browser will automatically open like before if we're interested

Dennull (approved):
Vite looks a lot faster, thanks for doing this!

src/index.tsx
NasihNazeem: Originally this was still a `.js` file instead of a `.tsx`. Once I changed the filetype to tsx it needed to know the type

vite.config.ts
NasihNazeem: Sure I can implement that

package.json
NasihNazeem: Hm I think I unintentionally removed this trying to follow some guides on stack overflow. If we had it originally, I'll bring it back

package.json
NasihNazeem: I have uninstalled react-scripts

src/index.tsx
Dennull: Oh ok makes sense, thanks!

PR #80, opened 2024-01-15
T75, T79: Update profile routes and provide placeholder picture and banner

### LINKED ISSUE #75: Provide placeholder avatar and banner

### LINKED ISSUE #79: Display other user profiles
Currently, the profile page only displays the current user's profile. To finish up the profile page feature, we want to update this to fetch other user's profiles based on the url. For example, the `/user1` route should fetch information about the person with the username `user1`

NasihNazeem:
This PR resolves #79 and resolves #75. This updates the profile page route to a variable that is a selected user's `username`. There has also been an update to the profile picture and banner to be more of a template rather than customizable. Lastly, we have updated the requests to send the username which as been passed as a parameter in the URL. This also introduces changes to provide routes from each `displayName` found on the app and any avatar. This is mainly to optimize for SEO so you can look up usernames instead of needing to remember user IDs 

Lastly, stylistic updates to the profile page include a border to the profile picture and divider updates to the `ExpandedPostReplies` and `PostItem`


src/components/Common/UserAvatar.tsx
```diff
@@ -0,0 +1,26 @@
+import { Avatar, Link } from "@mui/material";
+import React from "react";
```
Dennull: Think we can chuck this since it's not used

src/components/Common/FollowButton.tsx
```diff
@@ -4,6 +4,9 @@ const styles = {
   followButton: {
     boxShadow: "none",
     backgroundColor: "primary.main",
+    textTransform: "none",
+    fontWeight: "bold",
+    minWidth: "84px",
```
Dennull: Minor nitpick, can we sort these style objects alphabetically? I wish we could have eslint checks for this

src/components/Messages/ConversationListItem.tsx
```diff
@@ -39,11 +39,15 @@ const ConversationListItem = ({
   onClick,
   selected,
 }: ConversationListItemProps) => {
+  const handleAvatarClick = (e: React.MouseEvent) => {
+    e.stopPropagation();
+  };
+
```
Dennull: I think we can inline this since we're not doing that much and it's not reused

src/App.tsx
```diff
@@ -62,7 +62,10 @@ function App() {
         path="/messages/:userId1/:userId2"
         element={<ProtectedRoute component={DirectMessage} />}
       />
-      <Route path="/profile" element={<ProtectedRoute component={Profile} />} />
+      <Route
+        path="/user/:username"
```
Dennull: Do we need `/user/:username`? Or can we simplify it to `/:username`?

src/components/Common/UserAvatar.tsx
```diff
@@ -0,0 +1,26 @@
...
+  avatar: {
+    opacity: 0.75,
+    "&:hover": {
+      opacity: 1,
+    },
+  },
+};
+
+type UserAvatarProps = {
+  userId?: number;
```
Dennull: This should be a username

src/components/SideBar/RelevantUsers.tsx
Dennull, line 1: Looks like the instances of `userId` here should be changed to `username`.

src/pages/Profile.tsx
```diff
@@ -42,49 +65,62 @@ const styles = {
...
   tweetCount: { fontSize: 13 },
   username: {
     color: "#71797E",
-    fontSize: 16,
+    fontSize: 15,
   },
 };
 
 type ProfileContent = {
   postCount: number;
```
Dennull, line 93: Not really related, but do we need this? We can probably infer the post count by the length of the post list array. Technically it's a little risky since we're using the same store for all post lists, but in the case that we show the wrong posts we'd have bigger problems than the post count. Maybe we can make a ticket for this

src/pages/Profile.tsx
```diff
@@ -110,41 +147,37 @@ const Profile = () => {
...
+                    size="small"
+                    variant="outlined"
+                    startIcon={<EditIcon />}
+                    sx={styles.editProfileButton}
+                  >
+                    Edit Profile
+                  </Button>
+                ) : (
+                  <FollowButton />
+                )}
```
Dennull, line 173: Another kinda unrelated thing but the edit profile button shows for a brief moment when you first load the page. I think we already have a ticket for figuring out how to handle and `undefined` user, but we need loaders on a lot of things anyways so that can be another ticket

Dennull (changes requested):
It's looking good, just a few small comments!

src/components/Common/FollowButton.tsx
NasihNazeem: Sorted!

src/components/Messages/ConversationListItem.tsx
NasihNazeem: Alright so I've updated it to look like this instead

```
<ListItemAvatar
          onClick={(e: React.MouseEvent) => e.stopPropagation()}
          sx={styles.avatar}
>
          <UserAvatar userId={conversation.otherUserId} />
</ListItemAvatar>

File: src/App.tsx (hunk 4 above)
Comment (NasihNazeem) [reply src/App.tsx]: Yeah good point, I'll make this change

File: src/components/Common/UserAvatar.tsx (hunk 5 above)
Comment (NasihNazeem) [reply src/components/Common/UserAvatar.tsx]: Updated this, good eye

File: src/components/SideBar/RelevantUsers.tsx
Comment (NasihNazeem) [reply src/components/SideBar/RelevantUsers.tsx:1]: Updated this, thank you!

File: src/pages/Profile.tsx (hunk 6 above)
Comment (NasihNazeem) [reply src/pages/Profile.tsx:93]: Glad you brought this up because it just reminded me of something I was thinking about earlier.

I think we'll need the postCount variable instead of inferring it by the post list array, because at some point we'll have to include pagination to be able to infinite scroll and load more data and (I think that requires another fetch). So that postCount will be the count for all-time posts, but the post list will only be like the 25 recent posts.

Another thing I forgot to do in this ticket but I think I'll squeeze into this PR is changing of this data depending on what tab you are on. Similar to how Twitter does it depending on if you are looking at replies, posts or likes.

File: src/pages/Profile.tsx (hunk 7 above)
Comment (NasihNazeem) [reply src/pages/Profile.tsx:173]: Yeah definitely need loaders, I'll save this for that ticket

File: src/components/Messages/ConversationListItem.tsx (hunk 3 above)
Comment (Dennull) [reply src/components/Messages/ConversationListItem.tsx]: Nice!

File: src/pages/Profile.tsx (hunk 7 above)
Comment (Dennull) [reply src/pages/Profile.tsx:173]: Great, I made a draft ticket and threw it in the backlog

File: src/pages/Profile.tsx (hunk 6 above)
Comment (Dennull) [reply src/pages/Profile.tsx:93]: Yeah good point! I agree, let's leave this topic for when we tackle infinite scrolling

File: src/components/Posts/ExpandedPostItem.tsx [hunk 8]
```diff
@@ -113,14 +125,32 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
...
           <IconButton>
             <MoreVertIcon />
           </IconButton>
         }
-        title={`${post.displayName} `}
-        subheader={`@${post.username}`}
+        title={
+          <Link
+            component={Routerlink}
+            to={`/${post.userId}`}
```
Dennull: This should be `username`

Dennull (approved):

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: thank you!

PR #82, opened 2024-01-31
T81: Implemented follow/unfollow users

### LINKED ISSUE #81: Following other users
Users should be able to follow each other by clicking on the follow button on another user's profile. This button should toggle between follow/unfollow based on if the other user is currently being followed or not

NasihNazeem:
Resolves #81. This PR looks to introduce following/unfollowing functionality in all aspects of the application. This includes:

- Follow/Unfollow through profile
- Follow/Unfollow through expanded posts
- Follow/Unfollow through suggested users (not live as suggested users is dummy data)

In this PR, we introduce a few components:

- `FollowingButton.tsx` is where the styling and onClick for the following button is created.
- `FollowButton.tsx` is where the styling and onClick for the follow button is created.
- `UserButton.tsx` which acts as a parent component for `FollowButton.tsx` and `FollowingButton.tsx` where we handle the following/unfollowing business logic as well as fetching for the followStatus. 

In `Profile.tsx` we also update the `profileContents` variable to also retrieve the followStatus of the user when fetching for user data.

In `postsSlice.ts` we updated the expandedPosts state to implement followStatus as well as introducing a new reducer named `toggleFollow` to update the followStatus when following/unfollowing a user inside an expanded post.


src/components/Common/FollowButton.tsx
```diff
@@ -15,20 +15,15 @@ const styles = {
...
-  onClick?: () => void;
+  onClick: (e: React.MouseEvent) => void;
 };
 
 const FollowButton = ({ onClick }: FollowButtonProps) => {
-  const handleFollow = (e: React.MouseEvent) => {
-    e.stopPropagation();
-    onClick?.();
-  };
-
```
Dennull: Do you think we should move this stuff back in here to avoid having to write a `handleFollow` function every time we want to use the follow button? We can chuck the `onClick` to make it simpler to use and only bring it back if we have specific buttons that we want to perform specific actions

src/components/Common/FollowingButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+      backgroundColor: "gray.light",
+      color: "primary.main",
+      border: 1,
+      borderColor: "primary.main",
+    },
+  },
+};
+
+type FollowingButtonProps = {
+  onClick: (e: React.MouseEvent) => void;
```
Dennull: Similar to note with the `FollowButton`, I'd consider moving the common event handlers into the component

src/components/Common/UserButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+          },
+        }
+      );
+      setFollowStatus(result.data.followStatus);
+    };
+
+    if (user.username !== username) {
+      fetchFollowStatus();
+    }
+  }, [followStatus]);
```
Dennull: I'm a bit confused, is `initialFollowStatus` something that comes from the parent component? If it always comes from the parent component, do we need to fetch for it inside here?

vite.config.ts
```diff
@@ -8,6 +8,7 @@ export default defineConfig({
   // comment this out if that isn't relevant for your project
   build: {
     outDir: "build",
+    chunkSizeWarningLimit: 1000,
```
Dennull, line 11: For my own knowledge, what does this do?

src/state/slices/postsSlice.ts
```diff
@@ -104,6 +106,15 @@ export const postsSlice = createSlice({
...
     },
+    toggleFollow: (state, action: PayloadAction<boolean>) => {
+      return {
+        ...state,
+        expandedPost: {
+          ...state.expandedPost,
+          followStatus: !action.payload,
+        },
+      };
+    },
```
Dennull: Do you know if we have to return the entire state? Is it possible to mutate only the `expandedPost` object to keep it simple? Something like

```suggestion
    toggleFollow: (state, action: PayloadAction<boolean>) => {
      state.expandedPost.followStatus = !action.payload;
    },
```

Dennull (changes requested):

src/pages/Profile.tsx
```diff
@@ -97,13 +97,15 @@ type ProfileContent = {
   username: string;
   followerCount: number;
   followingCount: number;
+  followStatus: boolean;
```
Dennull, line 98: Another thing I noticed is that `followStatus` doesn't get updated right away when you click the follow/unfollow button. It only updates after a new fetch happens, so for a better experience we need to mutate the state in the front end as soon as the user clicks the button

Dennull (changes requested):

src/components/Common/UserButton.tsx
NasihNazeem: No need to fetch

src/state/slices/postsSlice.ts
NasihNazeem: Updated

src/components/Common/FollowButton.tsx
```diff
@@ -15,13 +18,32 @@ const styles = {
 };
 
 type FollowButtonProps = {
-  onClick?: () => void;
+  onClick: () => void;
+  username?: string;
```
Dennull: This relates to a comment I made on the backend, but should we swap to querying by the user's ID for consistency? I'm thinking we query by the username only on the initial GET request for their profile contents, then that GET request returns the id which we can use for subsequent queries

src/pages/Profile.tsx
```diff
@@ -112,30 +115,38 @@ const Profile = () => {
     username: "",
     followerCount: 0,
     followingCount: 0,
+    followStatus: false,
   });
 
   useEffect(() => {
+    setLoading(true);
```
Dennull, line 123: I don't think it makes a difference, but can we move the `setLoading` to inside of the `fetchProfileContents` function? I think it makes the order slightly more clear. So maybe do `setLoading(true)` before the try catch and do `setLoading(false)` in a `finally` block

src/pages/Profile.tsx
```diff
@@ -159,18 +170,37 @@ const Profile = () => {
...
-                    startIcon={<EditIcon />}
-                    sx={styles.editProfileButton}
-                  >
-                    Edit Profile
-                  </Button>
-                ) : (
-                  <FollowButton />
-                )}
+                {!loading &&
+                  (profileContents.username === user.username ? (
```
Dennull: If we do swap to using user ids change that here too

src/pages/Profile.tsx
```diff
@@ -159,18 +170,37 @@ const Profile = () => {
...
+                  ) : (
+                    <FollowButton
+                      onClick={() =>
+                        setProfileContents({
+                          ...profileContents,
+                          followStatus: true,
+                        })
+                      }
+                      username={username}
+                    />
```
Dennull: One other thing, we forgot to update the follower count after clicking the follow/unfollow button

Dennull (changes requested):

src/components/Common/FollowButton.tsx
NasihNazeem: Updated to use the userId of the user instead of the username for both follow and unfollow button

src/pages/Profile.tsx
NasihNazeem: Updated as described

src/pages/Profile.tsx
NasihNazeem: I needed to provide an initial number because it is a part of the profileContents object, and I decided to use `NaN` instead of making it undefined or null because that changes the type of userId from `number` to `number | undefined` or something.

I wanted to only render when userID is a non-zero number so this is the change:

```suggestion
              {!loading && profileContents.userId &&
                  (profileContents.userId === user.userId ? (
```

src/pages/Profile.tsx
NasihNazeem: Updated! but its a bit different to how I'd normally do it as: `followerCount: profileContents.followerCount + 1`  but this does not increment the count. Rather, it appends another 1 similar to a string, so a 1 follower account would show 11 followers, or 2 would show 21. 

I decided to use pre-incrementing because of the nature of setState since it wasn't updating as it should

NasihNazeem: We need to look into why `followerCount` changes type after the fetch. New ticket needed

src/pages/Profile.tsx
Dennull: Discussed this offline, we'll go with using `undefined` since that is a better descriptor for the situation of there not being any userId and holds up better under number checks

src/pages/Profile.tsx
Dennull: Yeah, we looked into this offline but couldn't find out why, so let's that works and explore this bug in another ticket

Dennull (approved):

PR #88, opened 2024-04-21
T87: Fix reply order bug

### LINKED ISSUE #87: Reply order glitch
Problem:

Currently replies are sorted by likes, but when posts have the same amount of likes there's no "tiebreaker" to sort them by, so it seems as if they are ordered randomly 

Image

Solution:

It was decided that we should continue sorting replies by likes then sort by timestamp (descending) after. So, if a post has the same amount of likes as another post, they will be sorted from most recent to least recent

Dennull:
Resolves #87. Replies are now sorted by likes and then timestamp

### DIFF
--- a/src/components/Posts/ExpandedPostReplies.tsx
+++ b/src/components/Posts/ExpandedPostReplies.tsx
@@ -27,7 +27,6 @@ const ExpandedPostReplies = ({ postId }: ExpandedPostRepliesProps) => {
           }
         );
         dispatch(setPosts(resultReplies.data as Post[]));
-        console.log(resultReplies);
       } catch (e) {
         console.log(e.message);
       }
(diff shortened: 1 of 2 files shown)


src/components/Posts/ExpandedPostReplies.tsx
```diff
@@ -22,22 +22,23 @@ const ExpandedPostReplies = ({ post }: ExpandedPostRepliesProps) => {
           {
             params: {
               userId: user.userId,
-              postId: post.postId,
+              postId: postId,
             },
           }
         );
         dispatch(setPosts(resultReplies.data as Post[]));
+        console.log(resultReplies);
```
NasihNazeem: Think we just need this console.log removed

NasihNazeem (approved):

src/components/Posts/ExpandedPostReplies.tsx
Dennull: Nice catch, thanks!

PR #89, opened 2024-04-26
T74: Edit profile modal

### LINKED ISSUE #74: Edit profile modal
Clicking on the edit profile button on a user's own profile page should open a modal where users can edit their profile details. Things that they can edit include:
- Display name
- Bio
- Birthday

Dennull:
Closes #74. This PR adds a modal that lets users edit their own profiles. The following profile details can be edited:

- Bio
- Birthday
- Display name

In addition, birthdays are now displayed on a user's profile page

image


src/components/Posts/RepliesModal.tsx
```diff
@@ -43,7 +43,7 @@ const styles = {
     maxHeight: "90%",
   },
   dialogContent: {
-    padding: 1,
+    "&.MuiDialogContent-root": { padding: 1 },
```
NasihNazeem: What's this for?

src/pages/Profile.tsx
```diff
@@ -132,154 +145,189 @@ const Profile = () => {
...
-      } finally {
+        console.log(error);
         setLoading(false);
       }
     };
     fetchProfileContents();
     window.scrollTo(0, 0);
   }, [username]);
 
+  // TODO: Make date utility functions or use a library for date formatting
```
NasihNazeem, line 164: Yeah I think this would be a great idea.

Here is a [nice library ](https://www.npmjs.com/package/dateformat) after a quick search that we could implement?

src/components/Profile/EditProfileModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+}: EditProfileModalProps) => {
+  const [bioValue, setBioValue] = useState(bio);
+  const [birthDateValue, setBirthDateValue] = useState(birthDate || new Date());
+  const [displayNameValue, setDisplayNameValue] = useState(displayName);
+  const user = useAppSelector((state) => state.user);
+
+  const saveProfile = async (e: React.SyntheticEvent) => {
+    e.preventDefault();
+    try {
+      const newBio = await axios.put("http://localhost:3001/api/profile", {
```
NasihNazeem: Can we rename this to `updatedProfileData`?

src/components/Profile/EditProfileModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+        </DialogTitle>
+        <DialogContent sx={styles.dialogContent}>
+          <Box sx={styles.textFieldContainer}>
+            <TextField
+              fullWidth
+              label="Display Name"
+              onChange={(e) => setDisplayNameValue(e.target.value)}
+              variant="outlined"
+              value={displayNameValue}
+              InputProps={{ sx: { borderRadius: 2 } }}
```
NasihNazeem: For consistency, let's add this styling to the `styles` object

src/components/Profile/EditProfileModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <Box sx={styles.textFieldContainer}>
+            <TextField
+              fullWidth
+              label="Bio"
+              onChange={(e) => setBioValue(e.target.value)}
+              multiline
+              rows={2}
+              variant="outlined"
+              value={bioValue}
+              InputProps={{ sx: { borderRadius: 2 } }}
```
NasihNazeem: Same comment as earlier about adding this to `styles` object

src/components/Profile/EditProfileModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+                    id="date"
+                    placeholder="Birth Date"
+                    variant="outlined"
+                    {...props}
+                  />
+                )}
+                label="Birth Date"
+                maxDate={new Date()}
+                onChange={(e) => {
+                  e && setBirthDateValue(e);
```
NasihNazeem: Is it possible for `e` to be null?

src/components/Profile/EditProfileModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+                }}
+                value={birthDateValue}
+              />
+            </LocalizationProvider>
+          </Box>
+        </DialogContent>
+        <DialogActions>
+          <Button type="submit" variant="outlined">
```
NasihNazeem, line 144: What do you think about centering this button? Or at least having the same padding as the form within the modal

src/components/Profile/EditProfileModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+  },
+  dialogContent: {
+    display: "flex",
+    flexDirection: "column",
+    padding: 0,
+  },
+  dialogTitle: {
+    paddingBottom: 0,
+    paddingX: 0.5,
+    paddingTop: 1,
```
NasihNazeem: Let's make the padding for Y: 1 so that it is even
```suggestion
    paddingX: 0.5,
    paddingY: 1,
```

src/components/Profile/EditProfileModal.tsx
```diff
@@ -0,0 +1,152 @@
...
+              fullWidth
+              label="Display Name"
+              onChange={(e) => setDisplayNameValue(e.target.value)}
+              variant="outlined"
+              value={displayNameValue}
+              InputProps={{ sx: { borderRadius: 2 } }}
+            />
+          </Box>
+          <Box sx={styles.textFieldContainer}>
+            <TextField
```
NasihNazeem, line 111: I've noticed this with a lot of our `TextField` components where we are getting this Uncaught TypeError when you type anything inside of a textfield. This is occurring in our other components as well.. Will need to look into it

src/pages/Profile.tsx
```diff
@@ -132,154 +145,189 @@ const Profile = () => {
...
-            <Box>
-              <Typography sx={styles.displayName}>
-                {profileContents.displayName}
-              </Typography>
-              <Typography sx={styles.tweetCount}>
-                {profileContents.postCount} Tweets
-              </Typography>
-            </Box>
-          </Box>
+        <>
```
NasihNazeem: Why don't we wrap the middleContent in a `Box`
```suggestion
        <Box>
            {...middle content...}
        </Box>
```

src/pages/Profile.tsx
```diff
@@ -132,154 +145,189 @@ const Profile = () => {
...
+              </Box>
+            )}
+          </Box>
+          {!loading && (
+            <EditProfileModal
+              bio={profileContents.bio}
+              birthDate={profileContents.birthDate}
+              displayName={profileContents.displayName}
+              onClose={() => setEditProfileModalOpen(false)}
+              onSubmit={(editedBio: EditableProfileContents) => {
```
NasihNazeem: Let's change the argument `editedBio` to something more general like `editedProfile`

src/pages/Profile.tsx
```diff
@@ -132,154 +145,189 @@ const Profile = () => {
...
+                  <Typography sx={styles.bio}>{profileContents.bio}</Typography>
+                )}
+                <Box sx={styles.personalInfo}>
+                  <Box sx={styles.personalInfoContent}>
+                    <CalendarMonthIcon />
+                    <Typography>
+                      Joined {formatJoinedDate(profileContents.joinedDate)}
+                    </Typography>
+                  </Box>
+                  {profileContents.birthDate && (
```
NasihNazeem, line 260: Great addition

src/pages/Profile.tsx
```diff
@@ -132,154 +145,189 @@ const Profile = () => {
...
+            </Box>
+            <Box>
+              <Box sx={styles.banner} />
+              <Box sx={styles.profileContent}>
+                <Box sx={styles.avatarContainer}>
+                  <Avatar sx={styles.avatar} />
+                  {!loading &&
+                    profileContents.userId &&
+                    (profileContents.userId === user.userId ? (
+                      <Button
```
NasihNazeem, line 209: Can we try changing the `hover` design? Maybe something like inverting the colours so that it becomes green background and white text?

NasihNazeem (changes requested):
Well done bro I left some suggestions down but the functionality works really well! Keep it up

src/pages/Profile.tsx
NasihNazeem, line 1: When updating the display name, we should look into updating the display name in `AccountMenu.tsx`, `Profile Replies.tsx`, `ProfileLikes.tsx`, `ProfilePosts.tsx` since you can see those components

src/components/Profile/EditProfileModal.tsx
NasihNazeem: Btw I think I only get this error when typing something in, and then trying to type again

src/components/Posts/RepliesModal.tsx
Dennull: I forgot, I think I just didn't like how much padding there is in the default styling. Let me know if you think removing it or keeping it looks better

src/pages/Profile.tsx
Dennull: Yeah, I agree, just seemed out of scope for this, so my vote it to make another ticket for it

src/components/Profile/EditProfileModal.tsx
Dennull: Yup

src/components/Profile/EditProfileModal.tsx
Dennull: Sure, I'll add the same padding

src/components/Profile/EditProfileModal.tsx
Dennull: Weird.. I wasn't able to reproduce it, but we should keep an eye out for this

src/pages/Profile.tsx
Dennull: I didn't wrap it in a `Box` because the `EditProfileModal` has to be there, and since modals don't really fit into the flow of the rest of the document I left it as a fragment.

With this being the case I ended up moving `EditProfileModal` outside of `Layout` entirely though because I think that it's not really part of the page. Let me know what you think about that

src/pages/Profile.tsx
Dennull: Yes

src/components/Profile/EditProfileModal.tsx
Dennull: Actually I ended up removing this. I realized we can just update the state on the front end using the value from the input fields, so no need to return anything from the backend

src/pages/Profile.tsx
Dennull: I personally like the subtle one we have more. Let's hop on a call and chat about it some time? And maybe get the rest of the team's opinion

src/components/Profile/EditProfileModal.tsx
Dennull: I think it's null if the user makes the input blank. Technically I guess that's ok cause people could remove their birthday if they wanted?

src/pages/Profile.tsx
Dennull: Yeah good catch! I ended up having to make new reducers in our user and post slices since the display names for those are stored in there. While implementing this, I ran into an issue where the entire app was rerendering whenever the display name changed. I found out it was because we were storing the entire `user` state inside of `App.tsx`

The better practice is to extract only the properties that we need from the state rather than returning the entire state. That way, the component only rerenders when that property changes. We do this in a few places, so we should try to optimize our Redux states better. I fell down a rabbit hole of Redux optimizations that could be made, but that's a lot of extra work, so I made a separate tech debt ticket for that

src/components/Profile/EditProfileModal.tsx
Dennull: As discussed offline, we'll deal with this validation type stuff in another ticket. I'll change `e` to `value` though, because that parameter is actually the value of the field, not the event object

NasihNazeem (approved):
Looking good!

PR #90, opened 2024-04-26
T86: Adding emojis

### LINKED ISSUE #86: Adding emojis
Functionality needs to be added to the emoji button on the text input fields. This [emoji picker library](https://www.npmjs.com/package/emoji-picker-react) might be a good place to look for a premade component to save time. The emoji icon button should open a dropdown where users can pick an emoji. Once they click on emoji on the dropdown, it should be put in the input field where their cursor is.

AnindaS13:
Used the `EmojiPicker` component from [emoji-picker-react](https://github.com/ealush/emoji-picker-react) to show emojis and let the user select them. This was added in three places:
- `DirectMessage.tsx`
- `ComposeReply.tsx`
- `ComposePost.tsx`

When the user clicks on the emoji icon, it will display the emoji picker, and when clicked on again it will close. If a user hits the post or send buttons, the emoji picker also disappears if it was open prior to hitting those buttons. The emoji picker component also closes if you click anywhere on the page outside of the component


src/components/Posts/ComposePost.tsx
```diff
@@ -79,19 +108,44 @@ const ComposePost = ({ placeholder, minRows }: ComposePostProps) => {
...
               </IconButton>
-              <IconButton size="small">
-                <EmojiEmotionsOutlinedIcon />
-              </IconButton>
+              <Stack direction="column">
+                <IconButton
+                  id="emoji-button"
+                  size="small"
+                  onClick={() => toggleEmojiPicker(false)}
+                >
```
NasihNazeem: I feel like updating the setState inline should suffice, you just need to toggle using the `!` to get the same functionality.

```suggestion
                  onClick={() => setShowEmojiPicker(!showEmojiPicker)}
```

src/components/Posts/ComposePost.tsx
```diff
@@ -28,13 +29,41 @@ const styles = {
...
 };
 
 const ComposePost = ({ placeholder, minRows }: ComposePostProps) => {
   const [postTextContent, setPostTextContent] = useState("");
+  const [showEmojiPicker, setShowEmojiPicker] = useState(false);
+  const emojiContainerRef = useRef<HTMLDivElement>(null);
   const user = useAppSelector((state) => state.user);
   const dispatch = useAppDispatch();
 
+  const toggleEmojiPicker = (isPostButton: Boolean) => {
```
NasihNazeem: What I'm understanding from this function is basically you are checking for whether this is a Post Button click or not. If it is, and the picker is showing, hide it. If the picker is not showing, do nothing. If this is not a post button click, hide the picker. 

I think you can simplify this because in the scenarios that matter we are only hiding the picker

src/components/Posts/ComposeReply.tsx
```diff
@@ -112,9 +132,30 @@ const ComposeReply = ({
...
                 <EmojiEmotionsOutlinedIcon />
               </IconButton>
+              {showEmojiPicker && (
+                <Box
+                  sx={styles.emojiContainer}
+                  ref={emojiContainerRef}
+                  tabIndex={0}
+                  onBlur={clickOffEmojiPicker}
+                >
+                  <EmojiPicker
```
NasihNazeem: This library also has a prop called `emojiStyle` where it is set default to value `apple`. If we want it to look more like twitter, we should set the value to `twitter`. The way I suggested it to be done does not have to be the only way. If you want to create a const for EmojiStyle.TWITTER and then use that, it'll look better as well. Up to you
```suggestion
                  <EmojiPicker
                    emojiStyle={EmojiStyle.TWITTER}
```

src/components/Posts/ComposePost.tsx
```diff
@@ -79,19 +108,44 @@ const ComposePost = ({ placeholder, minRows }: ComposePostProps) => {
...
           </Box>
+          {showEmojiPicker && (
+            <Box
+              sx={styles.emojiContainer}
+              ref={emojiContainerRef}
+              tabIndex={0}
+              onBlur={clickOffEmojiPicker}
+            >
+              <EmojiPicker
+                onEmojiClick={(emoji) => {
```
NasihNazeem: similar to comment in `ComposeReply.tsx`, emojiStyle can be changed

src/components/Posts/ComposeReply.tsx
```diff
@@ -74,6 +76,24 @@ const ComposeReply = ({
       console.log(err);
     }
   };
+  const [showEmojiPicker, setShowEmojiPicker] = useState(false);
+  const toggleEmojiPicker = () => {
```
NasihNazeem: inline setState works better I think..but tbh this is a design choice. If we want to say hey let's not do any inline, we should be consistent with the rest of our components.

src/components/Posts/ComposeReply.tsx
```diff
@@ -74,6 +76,24 @@ const ComposeReply = ({
...
+    setShowEmojiPicker(!showEmojiPicker);
+  };
+
+  const emojiContainerRef = useRef<HTMLDivElement>(null);
```
NasihNazeem: can we move this to the top of the component?

src/components/Posts/ComposeReply.tsx
```diff
@@ -112,9 +132,30 @@ const ComposeReply = ({
...
+                onClick={toggleEmojiPicker}
+                id="emoji-button"
+              >
                 <EmojiEmotionsOutlinedIcon />
               </IconButton>
+              {showEmojiPicker && (
+                <Box
+                  sx={styles.emojiContainer}
+                  ref={emojiContainerRef}
+                  tabIndex={0}
```
NasihNazeem: What does tabIndex do?

src/components/Posts/ComposeReply.tsx
NasihNazeem: After adjusting the paper props for the dialog component within `RepliesModal.tsx`, I think we should adjust the style of the picker. 

Here is what I recommend: 


```suggestion
                  <EmojiPicker
                    style={styles.emojiPicker as React.CSSProperties}
                    emojiStyle={EmojiStyle.TWITTER}
                    onEmojiClick={(emoji) => {
                      setPostTextContent(
                        (prevContent) => prevContent + emoji.emoji
                      );
                    }}
                    previewConfig={{ showPreview: false }}
                  />

```

look at styling comment for style adjustments.

src/components/Posts/ComposeReply.tsx
```diff
@@ -37,6 +38,7 @@ const styles = {
     padddingY: 0,
     justifyContent: "space-between",
   },
+  emojiContainer: { position: "absolute", zIndex: 1, marginTop: 5 },
 };
```
NasihNazeem, line 34: we should adjust the style of the picker:


```suggestion
  emojiPicker: {
    height: 300,
    width: 250,
    "--epr-emoji-size": "20px",
  },
```

NasihNazeem, line 34: adjust zIndex to be anything above Dialog zIndex in `RepliesModal.tsx`

src/components/Posts/ComposeReply.tsx
NasihNazeem, line 1: Adjust PaperProps styling of the Dialog component within `RepliesModal.tsx` and `PostButtonModal.tsx`

something like this:

/<Dialog
      fullWidth
      onClose={onClose}
      open={open}
      scroll="paper"
      PaperProps={{ style: styles.paperProps }}
      sx={styles.dialog}
    />

NasihNazeem: styles.paperProps should be something like

   where styles.paperProps is
    
      paperProps: {
    overflow: "visible",
    borderRadius: 20,
    },

src/pages/DirectMessage.tsx
NasihNazeem, line 1: please make similar adjustments to the EmojiPicker as mentioned in other files

NasihNazeem (changes requested):
Great work integrating the emoji picker into the application. It makes our app look more put together.

I had a few suggestions but if you need more explanation let's schedule something to talk about it

src/components/Posts/ComposePost.tsx
AnindaS13: The state is now set inline.

src/components/Posts/ComposePost.tsx
AnindaS13: Now we use the clickOffEmojiPicker() function to handle this for the post buttons. Toggle function has been completely removed.

src/components/Posts/ComposeReply.tsx
AnindaS13: I used the twitter emoji style.

src/components/Posts/ComposeReply.tsx
AnindaS13: Moved.

src/components/Posts/ComposeReply.tsx
AnindaS13: Implemented your suggested changes for this. I did not include the as React.CSSProperties part.

src/components/Posts/ComposeReply.tsx
AnindaS13: I will adjust the size a bit more

AnindaS13: Z-index is fine

AnindaS13: Adjusted the emoji size to 25 pixels.

Dennull (approved):

NasihNazeem (approved):

Issue #63, opened 2023-07-20
Navbar styling touch ups


christopherkinyua:
Our navbar could use a few styling improvements:

- It might look better if the text and icons of the navbar were bigger. Try experimenting with changing the font weight as well to make things stand out more
- Highlight the currently selected tab on navbar
- See if we can style the popup of the user avatar (where the logout button is) to look more like a modal (add a little arrow, maybe try adding some box shadow, etc). Also add an icon beside the "log out" text

PR #97, opened 2024-05-06
T91, T92, T96: Standardize typography

### LINKED ISSUE #91: Make typography consistent
Our typography is kind of all over the place right now, so we should take a moment to define our typography styles in our theme. Along with that, we should also discuss what type of fonts we should use where. For example, we should discuss what components we want to have using `h1`, `h2`, `body`, etc

### LINKED ISSUE #92: Bold all display names
User's display names should all be bolded. This adds a level of distinction between the username and display name, and will help add some color to the app. The places where this is relevant are:

- User avatar
- Posts
- Replies
- Conversation list
- Chat
- Profile page
- Suggested users

### LINKED ISSUE #96: Extra whitespace on really long posts
Posts with extremely long text have extra whitespace on them. I believe this might be because the container has a fixed width. Look into if we want to implement a character limit as well

Image

Image

Dennull:
This PR seeks to standardize some of the typography in the app. It closes #91, closes #92, and fixes #96. There are a lot of changes in this PR, but I essentially took a look at what different types of text we were using around the app and mapped them to specific
 variants in `Theme.tsx`. This way, we can just use MUI's `Typography` component and simply specify the `variant` prop whenever we wanted text to look a certain way.

The following variants were made:

- h1 - rarely used but I made it for larger titles like the welcome page or the "select a message" prompt
- h2 - used for the titles at the top of each page (eg. "Timeline" or "Messages")
- h3 - is mainly used for the navbar items, but is also used in a few other places in the app
- subtitle1 - the display name text, which is bolded
- subtitle2 - the username text, which is gray and has a lighter font weight
- body1 - the standard text used throughout the app
- body2 - the exact same as subtitle2 (the username text), but in a smaller font size. It's used for smaller captions like the footer or to display timestamps in the messages

image

### DIFF
--- a/src/components/Messages/ConversationList.tsx
+++ b/src/components/Messages/ConversationList.tsx
@@ -14,10 +14,12 @@ import CreateMessageModal from "./CreateMessageModal/CreateMessageModal";
 
 const styles = {
   header: {
+    alignItems: "center",
     display: "flex",
     justifyContent: "space-between",
-    paddingTop: 2,
-    paddingX: 2,
+    paddingTop: 1,
+    paddingLeft: 2,
+    paddingRight: 1,
   },
   searchBarContainer: { padding: 1 },
 };
@@ -46,7 +48,7 @@ const ConversationList = () => {
   return (
     <Box>
       <Box sx={styles.header}>
-        <Typography variant="h6">Messages</Typography>
+        <Typography variant="h2">Messages</Typography>
         <IconButton onClick={() => showMessageModal(true)}>
           <ChatOutlinedIcon />
         </IconButton>
--- a/src/components/Messages/ConversationListItem.tsx
+++ b/src/components/Messages/ConversationListItem.tsx
@@ -14,18 +14,23 @@ const styles = {
   avatar: { margin: "auto" },
   displayName: {
     flex: "1, 1, auto",
+    lineHeight: "inherit",
     minWidth: 0,
-    fontWeight: "bold",
   },
   primaryTextContainer: {
     display: "flex",
+    lineHeight: "1.25rem",
     gap: 0.5,
   },
   stack: {
     width: "100%",
   },
-  timestamp: { flex: "1, 0, auto", minWidth: "fit-content" },
-  username: { flex: "1, 1, auto", minWidth: 0 },
+  timestamp: {
+    flex: "1, 0, auto",
+    lineHeight: "inherit",
+    minWidth: "fit-content",
+  },
+  username: { flex: "1, 1, auto", lineHeight: "inherit", minWidth: 0 },
 };
 
 type ConversationListItemProps = {
@@ -52,19 +57,19 @@ const ConversationListItem = ({
           disableTypography
           primary={
             <Box sx={styles.primaryTextContainer}>
-              <Typography noWrap sx={styles.displayName} variant="body2">
+              <Typography noWrap sx={styles.displayName} variant="subtitle1">
                 {conversation.displayName}
               </Typography>
-              <Typography noWrap sx={styles.username} variant="body2">
+              <Typography noWrap sx={styles.username} variant="subtitle2">
                 {`@${conversation.username}`}
               </Typography>
-              <Typography noWrap sx={styles.timestamp} variant="body2">
+              <Typography noWrap sx={styles.timestamp} variant="subtitle2">
                 {`- ${formatTimestamp(conversation.timestamp)}`}
               </Typography>
             </Box>
           }
           secondary={
-            <Typography noWrap variant="body2">
+            <Typography noWrap variant="body1">
               {conversation.textContent}
             </Typography>
           }
--- a/src/components/Messages/ConversationListItem.tsx
+++ b/src/components/Messages/ConversationListItem.tsx
@@ -14,12 +14,10 @@ const styles = {
   avatar: { margin: "auto" },
   displayName: {
     flex: "1, 1, auto",
-    lineHeight: "inherit",
     minWidth: 0,
   },
   primaryTextContainer: {
     display: "flex",
-    lineHeight: "1.25rem",
     gap: 0.5,
   },
   stack: {
@@ -27,10 +25,10 @@ const styles = {
   },
   timestamp: {
     flex: "1, 0, auto",
-    lineHeight: "inherit",
+    fontSize: 15,
     minWidth: "fit-content",
   },
-  username: { flex: "1, 1, auto", lineHeight: "inherit", minWidth: 0 },
+  username: { flex: "1, 1, auto", fontSize: 15, minWidth: 0 },
 };
 
 type ConversationListItemProps = {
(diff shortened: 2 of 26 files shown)


src/components/Messages/ConversationList.tsx
```diff
@@ -14,10 +14,12 @@ import CreateMessageModal from "./CreateMessageModal/CreateMessageModal";
 
 const styles = {
   header: {
+    alignItems: "center",
     display: "flex",
     justifyContent: "space-between",
-    paddingTop: 2,
-    paddingX: 2,
+    paddingTop: 1,
```
NasihNazeem, line 20: I think we should keep the padding consistent with the Timeline header title where we do something like 

`padding: 2`

src/components/Messages/ConversationListItem.tsx
```diff
@@ -24,8 +23,12 @@ const styles = {
   stack: {
     width: "100%",
   },
-  timestamp: { flex: "1, 0, auto", minWidth: "fit-content" },
-  username: { flex: "1, 1, auto", minWidth: 0 },
+  timestamp: {
+    flex: "1, 0, auto",
+    fontSize: 15,
```
NasihNazeem, line 28: Let's remove this `fontSize`?

src/components/Messages/ConversationListItem.tsx
```diff
@@ -24,8 +23,12 @@ const styles = {
...
+    minWidth: "fit-content",
+  },
+  username: { flex: "1, 1, auto", fontSize: 15, minWidth: 0 },
```
NasihNazeem, line 31: Remove `fontSize` again I think

src/components/NavBar/AccountMenu.tsx
```diff
@@ -39,8 +39,13 @@ const AccountMenu = () => {
       <Button onClick={handleClick} sx={styles.button}>
         <Avatar />
         <Stack sx={styles.nameContainer}>
-          <Typography>{user.displayName}</Typography>
-          <Typography>{`@${user.username}`}</Typography>
+          <Typography color="primary" variant="subtitle1" fontSize={15}>
```
NasihNazeem: Do we still need this `fontSize`?

src/components/NavBar/NavItem.tsx
```diff
@@ -12,7 +12,10 @@ const NavItem = ({ icon, label, route }: NavItemProps) => {
   return (
     <ListItemButton component={Routerlink} to={route}>
       <ListItemIcon>{icon}</ListItemIcon>
-      <ListItemText primary={label} />
+      <ListItemText
+        primary={label}
+        primaryTypographyProps={{ variant: "h3" }}
```
NasihNazeem, line 17: We should figure out a way to have bold/not bold styling depending on whether this is selected (this should be done within [Navbar styling](https://github.com/Project-Chirp/chirp-frontend/issues/63))

src/components/SideBar/RelevantUsers.tsx
```diff
@@ -59,24 +48,27 @@ const RelevantUsers = () => {
...
-              sx={styles.username}
-            >{`@${relevantUser.username}`}</Typography>
+              variant="subtitle2"
+            >
+              {`@${relevantUser.username}`}
+            </Typography>
           }
           sx={styles.listItemText}
         />
+        {/* TODO: Remove this if the post is by the current user */}
```
NasihNazeem, line 71: Good catch!

src/components/SideBar/SidebarFooter.tsx
NasihNazeem, line 1: Nothing to add, but can we also remove the `Docs` link from the footer? I think we agreed to remove it in one of the meetings

src/pages/Profile.tsx
```diff
@@ -240,10 +223,10 @@ const Profile = () => {
                     ))}
                 </Box>
                 <Box sx={styles.nameContainer}>
-                  <Typography variant="h2" sx={styles.displayName}>
+                  <Typography variant="h3">
```
NasihNazeem: Can we make this more bold?

NasihNazeem (changes requested):
Mostly looks good, I think I just had a few comments but overall app layout is looking so good, well done bro genuinely

src/components/Messages/ConversationList.tsx
Dennull: The reason the padding is like this is cause I had to take into account the padding of the icon button in the header and the search bar under it. I can make the padding of the others 0 and compensate for it in the header, but at the end of the day it's kinda the same thing

If you'd prefer me to change the padding of the other components and make the header have a `padding: 2` I can, I don't have a strong opinion on it. Let me know if this makes sense or needs more explaining

src/components/Messages/ConversationListItem.tsx
Dennull: The display name in the title of the conversation has a font size of 15, so if the username and timestamp have a different font size to me it looks weird since they're right beside each other. I think this is a one-off situation so I'm ok with having it be a custom size. What do you think?

src/components/Messages/ConversationListItem.tsx
Dennull: See the comment above about the timestamp, the same applies here

src/components/NavBar/AccountMenu.tsx
Dennull: Nope, got rid of it!

src/components/NavBar/NavItem.tsx
Dennull: Yeah that would be cool. Maybe the font weight changes a bit. @christopherkinyua maybe take a look at this let us know if you can

src/components/SideBar/SidebarFooter.tsx
Dennull: Yup!

src/pages/Profile.tsx
Dennull: Do you think we should make this one use more bold or make all of the `h3` elements have a bolder font weight?

src/components/Messages/ConversationList.tsx
NasihNazeem: That makes sense, let's leave this as is

src/components/Messages/ConversationListItem.tsx
NasihNazeem: Okay yeah that's fair. I looked at some of the other components, and I don't see a need for it elsewhere

src/pages/Profile.tsx
NasihNazeem: I don't it is necessary in other areas we used `h3` as I like the boldness for the other h3...Just feel like this one feels a bit light

NasihNazeem (commented):
More feedback!

src/pages/Profile.tsx
Dennull: Ok, I'll make this one bolder

Dennull: So do you think I should I also make the title of this page the same boldness? The title of the profile page and the display name both use `h3`, so should they be the same font weight too?

NasihNazeem: I think the title is fine, just the display name in the user profile section. It's a nitpick but I think a slight increase in boldness will help the overall look

Dennull: Ok sounds good

NasihNazeem (approved):

PR #99, opened 2024-05-13
T84: Infinite scroll

### LINKED ISSUE #70: Bottom divider for PostItem missing
There is no divider separating each `PostItem` within `ExpandedPostReplies` so it looks weird, like one massive PostItem. Adding a `Divider` at the bottom of each `PostItem` should fix this

### LINKED ISSUE #84: Introducing infinite scroll to all profile page queries
The current queries return all existing posts in the profile page tabs and that slows down the application massively. In this issue, we will introduce an infinite scroll to the queries.

NasihNazeem:
Closes #84. Introduces infinite scroll using libraries such as `@tanstack/react-query` for infinite querying, and `react-infinite-scroll-component` for making scrollable components. A new hook, `useFetchPosts`, was introduced that takes a unique URL, and a designated userId to fetch data depending on the page. 

`PostLists` within the timeline, profile page and direct messages pages now have infinite scroll integrated This PR also resolves #70.


src/components/Messages/ConversationList.tsx
```diff
@@ -11,6 +11,10 @@ import {
 import ChatOutlinedIcon from "@mui/icons-material/ChatOutlined";
 import SearchBar from "../Common/SearchBar";
 import CreateMessageModal from "./CreateMessageModal/CreateMessageModal";
+import InfiniteScroll from "react-infinite-scroll-component";
+import { useInfiniteQuery } from "@tanstack/react-query";
+import { queryClient } from "../../utilities/queryClient";
+import PageLoader from "../../pages/PageLoader";
```
Dennull: These aren't used right? There are a few more unused imports in this PR too if you look at the linter on your VS Code. I think the linter will catch this stuff once I get ESLint set up

src/components/Posts/ExpandedPostReplies.tsx
```diff
@@ -1,48 +1,85 @@
...
-          <PostItem key={o.postId} post={o} />
+    <Stack divider={<Divider />}>
+      <InfiniteScroll
+        dataLength={posts.length}
+        next={fetchNextPage}
+        hasMore={hasNextPage}
+        loader={<PageLoader />}
+      >
+        {posts.map((o, index) => (
+          <Box key={index}>
```
Dennull: Is it possible to pass the `postId` as the key directly to the `PostItem` like how it was before? Same comment for other post lists too

src/components/Posts/ExpandedPostReplies.tsx
```diff
@@ -1,48 +1,85 @@
...
+  if (status === "pending") return <PageLoader />;
+  if (status === "error") return <Box>{error.message}</Box>; // TODO: Create an Error Component
 
   return (
-    <>
-      {posts
-        .filter((o) => o.parentPostId === postId)
-        .map((o) => (
-          <PostItem key={o.postId} post={o} />
+    <Stack divider={<Divider />}>
```
Dennull, line 17: Nice, this also resolves #70 right?

src/components/Posts/ExpandedPostReplies.tsx
```diff
@@ -1,48 +1,85 @@
...
-              userId: user.userId,
-              postId: postId,
-            },
-          }
-        );
+    queryClient.clear();
+    fetchReplies({ pageParam: 1 });
+  }, [postId]);
+
+  const fetchReplies = async ({ pageParam = 1 }) => {
```
Dennull: Is it possible to use your `useFetchPosts` hook for this?

src/pages/DirectMessage.tsx
```diff
@@ -108,25 +119,55 @@ const DirectMessage = () => {
...
       );
-    };
-    fetchDirectMessage();
-  }, [dispatch, userId1, userId2]);
+    } catch (error) {
+      console.log("error fetching user data:", error);
+    }
+  };
+
+  const fetchDirectMessage = async ({ pageParam = 1 }) => {
```
Dennull: Similar here, it'd be nice if we could use your `useFetchPosts` hook for this. Maybe we can take a stab at making things more generic so that all infinite scroll lists are using the same component and/or hooks

src/pages/DirectMessage.tsx
```diff
@@ -201,35 +246,52 @@ const DirectMessage = () => {
...
+            <List
+              component="div"
+              ref={messageRef}
+              sx={styles.messageList}
+              id={"scrollable"}
+            >
+              <InfiniteScroll
+                dataLength={messages.length}
+                next={fetchNextPage}
+                style={{ display: "flex", flexDirection: "column-reverse" }}
```
Dennull: Looks like you forgot to put this in a styles object

src/pages/DirectMessage.tsx
```diff
@@ -201,35 +246,52 @@ const DirectMessage = () => {
...
-                      </Typography>
-                    }
-                  />
-                </ListItem>
-              ))}
+            <List
+              component="div"
+              ref={messageRef}
+              sx={styles.messageList}
+              id={"scrollable"}
```
Dennull, line 218: Minor nitpick, can we make this id a little more specific? Maybe something like `directMessageList`? Also, for string props we don't need the curly braces
```suggestion
              id="scrollable"
```

src/pages/DirectMessage.tsx
```diff
@@ -201,35 +246,52 @@ const DirectMessage = () => {
...
+            >
+              <InfiniteScroll
+                dataLength={messages.length}
+                next={fetchNextPage}
+                style={{ display: "flex", flexDirection: "column-reverse" }}
+                inverse={true}
+                hasMore={hasNextPage}
+                loader={<PageLoader />}
+                scrollableTarget={"scrollable"}
```
Dennull, line 224: Same small nitpick here, we don't need the curly braces for string props
```suggestion
                scrollableTarget="scrollable"
```

src/utilities/queryClient.ts
Dennull, line 1: It seems overkill to have a whole file with only 3 lines. I think we might be able to export this from the `useFetchPosts` hook? Or if we do make a more generic component like I suggested in other places then we should put this there

src/pages/DirectMessage.tsx
```diff
@@ -108,25 +119,55 @@ const DirectMessage = () => {
...
+        setMessages(result.data as Message[]);
+      }
+
+      return result.data;
+    } catch (error) {
+      console.log("error fetching posts:", error);
+    }
+  };
+
+  const { error, status, hasNextPage, fetchNextPage } = useInfiniteQuery({
```
Dennull: All of this stuff seems to be repeated pretty often. It'd be nice if we could make a generic wrapper component that puts all of this infinite scroll logic in the same spot. I think we can put the `useEffect` and `useInfiniteQuery` calls that come with these infinite scroll lists into a component to avoid copy pasting this boilerplate

src/utilities/useFetchPosts.ts
Dennull, line 1: This is a great start, and like I alluded to in other comments I think we should take this hook and build off of it to abstract even more of the infinite scroll logic away. It'd be really cool if we could have a component called like `InfiniteScrollList` and use that everywhere. There is quite a bit to unpack that might be easier to do on a call, so let me know if you'd like that!

Dennull: One other thing, I noticed a weird bug on the lists that use infinite scroll. If you scroll so that more items load, swap to another tab then swap back again, the list seems to have to refetch for the items. You can visually see the new list items being loaded and replaced again, along with the scrollbar shifting, so it makes for a bad user experience

Dennull (changes requested):

src/components/Messages/ConversationList.tsx
NasihNazeem: Good catch!

src/components/Posts/ExpandedPostReplies.tsx
NasihNazeem: You right, idk why I changed this. I'll revert it

src/components/Posts/ExpandedPostReplies.tsx
NasihNazeem: Think so

src/pages/DirectMessage.tsx
NasihNazeem: Done

src/pages/DirectMessage.tsx
NasihNazeem: Done

src/pages/DirectMessage.tsx
NasihNazeem: I thought about doing this but the way I was implementing it doesn't really make sense to me. It all became a little too conditional, depending on the parameters do this or do that. Instead I chose to create another hook to maintain the Single Responsibility principle called `useFetchMessages`. 

To do this, I also looked into storing the messages data within the Redux store

src/pages/DirectMessage.tsx
NasihNazeem: Done

src/utilities/useFetchPosts.ts
NasihNazeem: I will look into this bug, but have not found a solution just yet.

src/components/Posts/ExpandedPostReplies.tsx
NasihNazeem: Updated useFetchPosts to work on ExpandedPostReplies!

src/utilities/queryClient.ts
NasihNazeem: Moved this functionality into the generic component `InfiniteScrollList.tsx`

src/pages/DirectMessage.tsx
NasihNazeem: Done!

src/utilities/useFetchPosts.ts
NasihNazeem: We changed the default value to false for window re-fetching within the `QueryClient` as it was re-fetching whenever we refocused on our application.

Dennull:
As discussed offline, we'll convert this back to a draft while we figure out the best way to move forward. I think there's some confusion between how we're managing client side and server side state. If we want to introduce the react query library into the app we might want to change how we're managing our server side state in general, so we should probably fully understand that before implementing infinite scroll

PR #100, opened 2024-05-19
T86: Alternate implementation of emoji picker

Dennull:
This is a proposed alternative implementation of the emoji picker for #86. This implementation builds off of an existing PR, #90. Some changes that were made here compared to that version are:

- Including the icon button for users to open the emoji picker inside of the emoji picker component
- Using a ref attached to the icon button to ignore the onClick that closes the emoji picker rather than a button id
- The addition of extra props to the emoji picker component for customization such as width, height, and position
- Some general code cleanup


src/components/Common/EmojiPickerIconButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+}: EmojiPickerIconButtonProps) => {
+  const emojiContainerRef = useRef<HTMLDivElement>(null);
+  const emojiIconButtonRef = useRef<HTMLButtonElement>(null);
+  const twitterEmojiStyle = EmojiStyle.TWITTER;
+  const [showEmojiPicker, setShowEmojiPicker] = useState(false);
+  const top = pickerPosition === "bottom" ? 150 : -460;
+
+  const handleClick = (event: MouseEvent) => {
+    if (
+      emojiContainerRef.current &&
```
NasihNazeem, line 34: using the two refs is good! didnt know how to implement this so this was the right idea

src/components/Common/EmojiPickerIconButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+  pickerPosition = "bottom",
+  pickerWidth = 350,
+}: EmojiPickerIconButtonProps) => {
+  const emojiContainerRef = useRef<HTMLDivElement>(null);
+  const emojiIconButtonRef = useRef<HTMLButtonElement>(null);
+  const twitterEmojiStyle = EmojiStyle.TWITTER;
+  const [showEmojiPicker, setShowEmojiPicker] = useState(false);
+  const top = pickerPosition === "bottom" ? 150 : -460;
+
+  const handleClick = (event: MouseEvent) => {
```
NasihNazeem, line 35: Using the two refs is good! didnt know how to implement this so this was the right idea

src/components/Common/EmojiPickerIconButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <EmojiEmotionsOutlinedIcon />
+      </IconButton>
+      <Box sx={{ ...styles.emojiContainer, top: top }} ref={emojiContainerRef}>
+        <EmojiPicker
+          emojiStyle={twitterEmojiStyle}
+          height={pickerHeight}
+          onEmojiClick={onEmojiClick}
+          open={showEmojiPicker}
+          previewConfig={{ showPreview: false }}
+          style={styles.emojiPicker as React.CSSProperties}
```
NasihNazeem, line 70: Think we need to find a solution for this, if we need to look into doing this in other components I suggest we change how we define our styles object

src/components/Common/EmojiPickerIconButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }, []);
+
+  return (
+    <>
+      <IconButton
+        ref={emojiIconButtonRef}
+        size="small"
+        onClick={() => setShowEmojiPicker(!showEmojiPicker)}
+      >
+        <EmojiEmotionsOutlinedIcon />
```
NasihNazeem, line 61: Would be cool if we kept the emoji icon green when emoji picker is active, and grey when inactive

src/components/Common/EmojiPickerIconButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+  pickerHeight = 450,
+  pickerPosition = "bottom",
+  pickerWidth = 350,
+}: EmojiPickerIconButtonProps) => {
+  const emojiContainerRef = useRef<HTMLDivElement>(null);
+  const emojiIconButtonRef = useRef<HTMLButtonElement>(null);
+  const twitterEmojiStyle = EmojiStyle.TWITTER;
+  const [showEmojiPicker, setShowEmojiPicker] = useState(false);
+  const top = pickerPosition === "bottom" ? 150 : -460;
+
```
NasihNazeem, line 34: If we follow my suggestion, top position would need to go down by about 100..

```suggestion
  const top = topPosition ? -360 : 150;
```

src/components/Common/EmojiPickerIconButton.tsx
```diff
@@ -0,0 +1,74 @@
+import { Box, IconButton } from "@mui/material";
+import EmojiEmotionsOutlinedIcon from "@mui/icons-material/EmojiEmotionsOutlined";
+import { useState, useEffect, useRef } from "react";
+import EmojiPicker, { EmojiClickData, EmojiStyle } from "emoji-picker-react";
+
+type EmojiPickerIconButtonProps = {
+  onEmojiClick: (emoji: EmojiClickData) => void;
+  pickerHeight?: number;
+  pickerPosition?: "top" | "bottom";
+  pickerWidth?: number;
```
NasihNazeem, line 9: If this is binary, lets just have an optional prop like `topPosition` where we we keep it as a boolean, and default top to be false and use that instead. To get it to turn true, we just need to mention the prop in `DirectMessage.tsx`

src/components/Common/EmojiPickerIconButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+    "--epr-emoji-size": "25px",
+  },
+};
+
+const EmojiPickerIconButton = ({
+  onEmojiClick,
+  pickerHeight = 450,
+  pickerPosition = "bottom",
+  pickerWidth = 350,
+}: EmojiPickerIconButtonProps) => {
```
NasihNazeem, line 28: Why is `pickerHeight` and `pickerWidth` a prop?

Can we reduce it to something like 350 and 300 respectively

NasihNazeem, line 28: If we change it to a boolean, just need to default to false

src/components/Common/EmojiPickerIconButton.tsx
NasihNazeem: This isn't the ideal solution as we are hardcoding the numbers, but I think I want to prioritize shipping this part and worrying about dynamic pages at a later date

src/pages/DirectMessage.tsx
```diff
@@ -252,25 +247,14 @@ const DirectMessage = () => {
...
-                            emojiContainerStyle={styles.emojiContainer}
-                          />
-                        )}
+                        <EmojiPickerIconButton
+                          onEmojiClick={(emoji: EmojiClickData) => {
+                            setTextContent(
+                              (prevContent) => prevContent + emoji.emoji
+                            );
+                          }}
+                          pickerPosition="top"
```
NasihNazeem: If we take my suggestion, this just becomes `topPosition` or whatever we call the boolean

NasihNazeem (commented):
Huge job figuring out how to get the event listeners to work. I guess it was important to get the two refs into the same page

NasihNazeem (changes requested):
Some feedback for you, overall great job bro looks great

src/components/Common/EmojiPickerIconButton.tsx
Dennull: Yeah I think it's because to modify their emoji picker component we have to use a custom css property called "--epr-emoji-size". Since this is a custom property it's not defined in the `React.CSSProperties` type, so typescript complains

I think the only alternative solution would be to define a custom interface that extends the `React.CSSProperties` like so:

```
interface EmojiPickerCSSProperties extends React.CSSProperties {
  "--epr-emoji-size"?: string;
}

const styles = {
  emojiContainer: { position: "absolute", zIndex: 1 },
};

const emojiPickerStyle: EmojiPickerCSSProperties = {
  "--epr-emoji-size": "25px",
};

...

<EmojiPicker
  ...
  style={emojiPickerStyle}
 />
```

What do you think? I don't have a strong opinion on whether we make a custom interface or cast the type, so I'm good to implement whichever we want

src/components/Common/EmojiPickerIconButton.tsx
Dennull: Yea sure, I can do that. It sucks a little because we defined the color for our icon buttons in the theme to be gray, so we can't use the `color` prop of the icon button. To get around it we have to use the `sx` prop and do some conditional stuff, which is why there's some funky syntax:

```
const styles = {
...
  activeButton: {
    color: "primary.main",
  },
};

...

<IconButton
  ref={emojiIconButtonRef}
  onClick={() => setShowEmojiPicker(!showEmojiPicker)}
  size="small"
  sx={{ ...(showEmojiPicker && styles.activeButton) }}
 >
  <EmojiEmotionsOutlinedIcon />
</IconButton>

```

src/components/Common/EmojiPickerIconButton.tsx
Dennull: Yea I think this is a better suggestion, thanks!

src/components/Common/EmojiPickerIconButton.tsx
Dennull: I made them props so we can choose the height and width if we'd like. Maybe not necessary but I thought it'd be a nice to have the option. I'll change it to 350 and 300

src/components/Common/EmojiPickerIconButton.tsx
Dennull: I changed the code as follows: 

`const top = topPosition ? -(pickerHeight + 10) : 150;`

I think this is fine because this way the picker will always be 10 pixels above the button if it's positioned on top. And if it's positioned on the bottom I think hard coding it to 150 will always make it 10 pixels below the button (unless the height of the picker is <150, but that would be a terribly small emoji picker)

src/components/Common/EmojiPickerIconButton.tsx
NasihNazeem: Not gonna lie, I think let's leave it as you originally had it..maybe we can abstract the interface away and just use that for every styles object in our app..but that sounds unnecessary unless this happens occurs more often

src/components/Common/EmojiPickerIconButton.tsx
NasihNazeem: Ah okay that makes sense. Let's keep them as props in that case

src/components/Common/EmojiPickerIconButton.tsx
NasihNazeem: Yeah thats fair enough, good stuff dennis

src/components/Common/EmojiPickerIconButton.tsx
NasihNazeem: Damn I forgot about that, what do you think? worth it? I think having it green is good for UX

NasihNazeem (approved):
I think its good to merge into the other branch, good job bro

src/components/Common/EmojiPickerIconButton.tsx
Dennull: Aight sounds good to me!

src/components/Common/EmojiPickerIconButton.tsx
Dennull: Yeah I agree I think it looks nice. I'm fine with the syntax, it's a bit confusing at first but it makes sense to me and this case is sort of a one off so I think it's whatever

PR #101, opened 2024-05-25
T63: Navbar styling

### LINKED ISSUE #63: Navbar styling touch ups

christopherkinyua:
Finally did some changes to the navbar styling which include:
 - Added a switch to alternative icons and text bolding when a navbar item is selected
 - Made styling changes to the popover model for the log out
 - Added a selected highlight for the navbar


src/components/NavBar/AccountMenu.tsx
```diff
@@ -1,18 +1,35 @@
...
 const styles = {
   button: {
     marginTop: "auto",
     marginBottom: 2,
   },
   nameContainer: { paddingLeft: 2, textAlign: "left" },
+  popover: {
+    paddingTop: "10px",
+    paddingBottom: "10px",
+    borderRadius: "10px",
```
Dennull: Can we please make these numbers so they match with our theme spacing variables? Basically, any time you specify a raw number in the `sx` prop, that number gets multiplied by the theme spacing variable ((in our case 8 pixels) and that's the amount of pixels that gets set for that property.

For example, if you do `padding: 1`, that means a padding of 1 * 8 pixels = 8 pixels. If you do `padding: 2`, that means a padding of 2 * 8 pixels = 16 pixels. This is done so that we have consistent spacing units across the app. I'd suggest something like this:
```suggestion
    padding: 1,
    borderRadius: 3,
```

src/components/NavBar/AccountMenu.tsx
```diff
@@ -1,18 +1,35 @@
...
+  },
+  popoverButton: {
+    paddingTop: "5px",
+    paddingBottom: "5px",
+    paddingRight: "20px",
+    paddingLeft: "20px",
```
Dennull: Same here, please use numbers. If we also shift some of the padding to the pop over component we can reduce the padding on the buttons. Also a cool thing with MUI is that they give us shortcuts for things like adding padding only to the horizonal axis; we can use `paddingX` and `paddingY` to add padding to only the X axis and Y axis respectively (the same applies to margin). If we follow the padding suggestion for the popover I mentioned earlier I think we can do something like:
```suggestion
    paddingX: 2
```

src/components/NavBar/AccountMenu.tsx
```diff
@@ -1,18 +1,35 @@
...
+    textTransform: "none",
```
Dennull: This can be removed, we've already applied this in our theme

src/components/NavBar/AccountMenu.tsx
```diff
@@ -1,18 +1,35 @@
...
+  },
+  popoverIcon: {
+    marginRight: "10px",
```
Dennull: Same here, can we make this 
```suggestion
    marginRight: 1.5,
```

src/components/NavBar/AccountMenu.tsx
```diff
@@ -1,18 +1,35 @@
...
+    borderRadius: "10px",
+  },
+  popoverButton: {
+    paddingTop: "5px",
+    paddingBottom: "5px",
+    paddingRight: "20px",
+    paddingLeft: "20px",
+    textTransform: "none",
+  },
+  popoverIcon: {
```
Dennull: Nitpick, can we please rename this to something like `logOutIcon`?

src/components/NavBar/AccountMenu.tsx
```diff
@@ -61,8 +78,14 @@ const AccountMenu = () => {
           vertical: "bottom",
           horizontal: "center",
         }}
+        sx={{ ".MuiPopover-paper": styles.popover }}
```
Dennull: If we move the selector to the styles object I think we can keep the syntax consistent with the rest of the app

```
const styles = {
  popover: {
    ".MuiPopover-paper": {
      padding: 1,
      borderRadius: 3,
    },
  },
};

...

<Popover>
  ...
  sx={styles.popover}
</Popover>
```

src/components/NavBar/NavBar.tsx
```diff
@@ -1,13 +1,17 @@
...
 import AccountMenu from "./AccountMenu";
 import NavItem from "./NavItem";
 import PostButtonModal from "./PostButtonModal";
 import { useState } from "react";
 import ComposePost from "../Posts/ComposePost";
 import { useAppSelector } from "../../state/hooks";
+import { useLocation } from "react-router-dom";
 
 const styles = {
   icon: {
```
Dennull, line 19: The font size of the icons seems a little too big to me, do you think adding `fontSize: "30px"` here would make it look better?

src/components/NavBar/NavBar.tsx
```diff
@@ -33,33 +37,42 @@ const styles = {
...
+  const location = useLocation();
   const [openModal, setOpenModal] = useState(false);
   const { selectedConversation } = useAppSelector((state) => state.messages);
   const user = useAppSelector((state) => state.user);
 
   const navItems = [
     {
-      icon: <HomeIcon sx={styles.icon} />,
+      icon: <HomeOutlinedIcon fontSize="large" sx={styles.icon} />,
+      altIcon: <HomeIcon fontSize="large" sx={styles.icon} />,
```
Dennull: Minor nitpick, can we please rename this to `selectedIcon`?

src/components/NavBar/NavBar.tsx
```diff
@@ -33,33 +37,42 @@ const styles = {
...
 const NavBar = () => {
+  const location = useLocation();
   const [openModal, setOpenModal] = useState(false);
   const { selectedConversation } = useAppSelector((state) => state.messages);
   const user = useAppSelector((state) => state.user);
 
   const navItems = [
     {
-      icon: <HomeIcon sx={styles.icon} />,
+      icon: <HomeOutlinedIcon fontSize="large" sx={styles.icon} />,
```
Dennull: If we were to go with my suggestion of adding the font size in the `sx` object, we can delete the `fontSize` prop

src/components/NavBar/NavBar.tsx
```diff
@@ -33,33 +37,42 @@ const styles = {
...
     {
-      icon: <AccountCircleIcon sx={styles.icon} />,
+      icon: <AccountCircleOutlinedIcon fontSize="large" sx={styles.icon} />,
+      altIcon: <AccountCircleIcon fontSize="large" sx={styles.icon} />,
       label: "Profile",
       route: `/${user.username}`,
     },
   ];
-
+  console.log(location);
```
Dennull: Forgot to delete this

src/components/NavBar/NavBar.tsx
```diff
@@ -69,13 +82,15 @@ const NavBar = () => {
...
-          <List component="nav">
+          <List sx={styles.list} component="nav">
             {navItems.map((navItem, index) => (
               <NavItem
                 key={index}
                 icon={navItem.icon}
+                altIcon={navItem.altIcon}
                 label={navItem.label}
                 route={navItem.route}
+                active={location.pathname === navItem.route}
```
Dennull: I kind of think `selected` is a more descriptive name, what do you think?

src/components/NavBar/NavItem.tsx
Dennull, line 1: I'd like to make some small styling tweaks to how the text of the nav items look, but I think it'll be easier if I make a PR to show you after. I'll include it in my changes to the account menu popover to add that little arrow

Dennull (changes requested):

src/components/NavBar/AccountMenu.tsx
christopherkinyua: Updated as per your suggestions mentioned above

src/components/NavBar/AccountMenu.tsx
christopherkinyua: Thanks for this, I see a bunch more suggestions and shortcuts in the MUI spacing documentation. Regarding the suggestions, I have updated it.

src/components/NavBar/AccountMenu.tsx
christopherkinyua: Done

src/components/NavBar/AccountMenu.tsx
christopherkinyua: Okay, I'll remove this

src/components/NavBar/AccountMenu.tsx
christopherkinyua: Done

src/components/NavBar/NavBar.tsx
christopherkinyua: Yes, it is. I have updated the change.

src/components/NavBar/NavBar.tsx
christopherkinyua: Deleted

src/components/NavBar/NavBar.tsx
christopherkinyua: Done

src/components/NavBar/AccountMenu.tsx
christopherkinyua: Done

src/components/NavBar/NavBar.tsx
christopherkinyua: Done

src/components/NavBar/NavBar.tsx
christopherkinyua: Updated

Dennull (approved):
Good job, I think this is a small quality of life thing that makes a big difference on the overall appearance of the app!

PR #103, opened 2024-05-27
T102: Fix emoji picker overflow bug

### LINKED ISSUE #102: Emoji picker overflow is scroll in modals
There's a bug with the emoji picker when it's used in the `RepliesModal` and the `PostModal`. The emoji picker overflows the modal, and since the overflow isn't visible, it turns the modal scrollable. This sucks because the entire modal is only a certain height so you can't see the entire post text field and emoji picker at the same time

Image

Dennull:
This resolves #102.

**Context**

There is a bug with the emoji picker when it's used in our modals, specifically `PostButtonModal` and `RepliesModal`. The bug makes the emoji picker overflow so that the entire modal isn't visible (ie. it has a scroll bar)

**Solution**

The solution is to change the overflow to visible in the dialog content div. Additional refactors were also made to get rid of an MUI `Card` component that was being used in the `RepliesModal` because the styling of the card when overflow was visible made the modal look different

### DIFF
--- a/src/components/Common/EmojiPickerIconButton.tsx
+++ b/src/components/Common/EmojiPickerIconButton.tsx
@@ -11,7 +11,7 @@ type EmojiPickerIconButtonProps = {
 };
 
 const styles = {
-  emojiContainer: { position: "absolute", zIndex: 1 },
+  emojiContainer: { position: "absolute", zIndex: 1, overflow: "visible" },
   emojiPicker: {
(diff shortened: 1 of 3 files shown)


NasihNazeem (approved):

NasihNazeem:
Don't really have anything to critique with these changes, works as expected. Good work bro

PR #105, opened 2024-06-11
T71: Close post button modal after making a new post

### LINKED ISSUE #71: PostButtonModal onClose not working
The `onClose` for `PostButtonModal` with the Navbar is not working when you submit a new post and you currently have to close the modal manually instead. When creating a new post, we should be closing the modal on the button click

christopherkinyua:
I managed to address the issue of the post-modal button not closing. I added an `onClose` prop to the `ComposePost` and passed the `setOpenModal` function into it to invoke the function after a post is made. I made it conditional to check if `onClose` is provided. I was thinking about other cases when the `ComposePost`is called but no `onClose` is needed. Anyway, PR is ready for your review

### DIFF
--- a/src/components/Posts/ComposePost.tsx
+++ b/src/components/Posts/ComposePost.tsx
@@ -53,7 +53,9 @@ const ComposePost = ({ placeholder, onClose }: ComposePostProps) => {
           displayName: user.displayName,
         })
       );
-      onClose();
+      if (onClose) {
+        onClose();
+      }
     } catch (err) {
       console.log(err);
     }
--- a/src/components/Posts/ComposePost.tsx
+++ b/src/components/Posts/ComposePost.tsx
@@ -10,6 +10,7 @@ import EmojiPickerIconButton from "../Common/EmojiPickerIconButton";
 
 type ComposePostProps = {
   placeholder: string;
+  onClose: () => void;
 };
 
 const styles = {
@@ -31,7 +32,7 @@ const styles = {
   },
 };
 
-const ComposePost = ({ placeholder }: ComposePostProps) => {
+const ComposePost = ({ placeholder, onClose }: ComposePostProps) => {
   const [postTextContent, setPostTextContent] = useState("");
   const user = useAppSelector((state) => state.user);
   const dispatch = useAppDispatch();
@@ -52,6 +53,7 @@ const ComposePost = ({ placeholder }: ComposePostProps) => {
           displayName: user.displayName,
         })
       );
+      onClose();
     } catch (err) {
       console.log(err);
     }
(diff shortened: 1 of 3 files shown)


src/components/Posts/ComposePost.tsx
```diff
@@ -10,6 +10,7 @@ import EmojiPickerIconButton from "../Common/EmojiPickerIconButton";
 
 type ComposePostProps = {
   placeholder: string;
+  onClose: () => void;
```
Dennull: Can we make this an optional prop instead? Like this
```suggestion
  onClose?: () => void;
```

src/components/Posts/ComposePost.tsx
```diff
@@ -52,6 +53,9 @@ const ComposePost = ({ placeholder }: ComposePostProps) => {
           displayName: user.displayName,
         })
       );
+      if (onClose) {
+        onClose();
+      }
```
Dennull: If we make the prop optional you're right that we need an if check. However, we can simplify it using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining). This basically checks to see if the function is defined before running it, so it achieves the same thing as your if statement
```suggestion
      onClose?.()
```

src/pages/Timeline.tsx
```diff
@@ -18,7 +18,7 @@ const Timeline = () => {
           <Typography sx={styles.headerTitle} variant="h2">
             Timeline
           </Typography>
-          <ComposePost placeholder="What's happening?" />
+          <ComposePost placeholder="What's happening?" onClose={() => {}} />
```
Dennull: If we make `onClose` an optional prop, we can delete this
```suggestion
          <ComposePost placeholder="What's happening?"/>
```

Dennull (changes requested):

src/components/Posts/ComposePost.tsx
christopherkinyua: Nice, so just adding a question mark before make it conditional? I have implemented the change

src/components/Posts/ComposePost.tsx
christopherkinyua: Done done done

src/pages/Timeline.tsx
christopherkinyua: Alright Cap, done

src/components/Posts/ComposePost.tsx
Dennull: Nice. And yes it makes an optional type. It's basically the same as `() => void | undefined`, so the onClose property is allowed to be undefined

Dennull (approved):

PR #107, opened 2024-06-27
T85: Implement user search bar

### LINKED ISSUE #85: App search bar dropdown
The search bar should display a dropdown with the latest search results when users type. The MVP requirements are:

- When the user types a dropdown should appear showing the top 5 search results. A "view more" button at the bottom of the dropdown should appear so users can see more results if they want
  - Limit to 15 results total
  - The dropdown should be scrollable, so it should have a maximum size and scroll after that

NasihNazeem:
## Related Issues

Closes #85 

## Summary

This pull request introduces a fully functional search bar that allows users to search for other users within the application by username or display name. The search results are ordered by the number of followers, from most to least. Each result displays a temporary avatar, the username, and the display name of the user

This enhancement aims to improve user experience by making it easier and faster to find other users within the app

## Changes Made

- Debouncing: Improves search performance by limiting the rate of search queries.
- Clear Search: Provides a clear button to reset the search input.
- Blur on Select: Automatically blurs the search input when a user is selected from the results.

## Screenshots

image

## Testing Instructions

The template is still up for debate so please do suggest any changes you think are appropriate


package-lock.json
Dennull, line 1: Can I ask what got updated here? Was it an `npm audit fix`?

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+  const debouncedSetter = useMemo(
+    () => debounce((keywords: string) => setKeywords(keywords), 20),
+    []
+  );
+  const [focusSearchBar, setFocusSearchBar] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [keywords, setKeywords] = useState("");
+  const [loading, setLoading] = useState(false);
+  const [userList, setUserList] = useState<SelectedUser[]>([]);
+  const { getAccessTokenSilently } = useAuth0();
```
Dennull: How come we don't need to use this in other places? Is it because the jwt request hook hasn't been implemented yet?

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+      <Autocomplete
+        disablePortal
+        forcePopupIcon={false}
+        fullWidth
+        filterOptions={(x) => x}
+        getOptionLabel={(option) => `${option.displayName} @${option.username}`}
+        id="search"
+        inputValue={keywords}
+        ListboxProps={{ sx: styles.listBox }}
+        loading={loading}
```
Dennull: This `loading` state stays true if there's nothing typed in the input. The code path for an empty input returns right away without ever setting the state to false which makes the "Loading..." text to display when the user hasn't typed anything

On another note, something to look into would be if we could put a separate prompt for when the user hasn't typed anything in yet. For example, something like "Try searching for people". Similarly, displaying a "No results found" message when the backend returns no options would be nice too

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
 
 type SearchBarProps = {
   placeholder: string;
 };
 
+const styles = {
+  autocomplete: {
+    "&.MuiAutocomplete-input": { paddingX: 0 },
+    position: "relative",
+  },
```
Dennull: I'm not sure these do anything do they?

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+              onClick={() => onSelect(option.username)}
+            >
+              <ListItemAvatar>
+                <Avatar />
+              </ListItemAvatar>
+              <ListItemText
+                disableTypography
+                primary={
+                  <Box>
+                    <Typography sx={styles.displayName} variant="body1">
```
Dennull: Can we use the `subtitle1` variant? We use that for display names everywhere, so we can get rid of the style. See what our typography variants are used for in #91

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+                      {option.displayName}
+                    </Typography>
+                    <Typography variant="body2">{`@${option.username}`}</Typography>
```
Dennull: Similar here, let's use `subtitle2` for the username

src/components/Common/SearchBar.tsx
Dennull, line 1: This component is actually used in other places like the conversation list, and it looks different there. Maybe we should create a new search bar component, one with a dropdown and one without a dropdown

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+
+  const handleClear = () => {
+    setKeywords("");
+    setUserList([]);
+  };
+
+  const handleInputChange = (newInputValue: string) => {
+    debouncedSetter(newInputValue);
+    setFocusSearchBar(true);
+    if (newInputValue === "") {
```
Dennull: Should we use `.trim()` here for white spaces?

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+      setUserList([]);
+    }
+  };
+
+  const onSelect = (selectedUsername: string) => {
+    setFocusSearchBar(false);
+    setKeywords("");
+    inputRef.current?.blur();
+    const path = `/${selectedUsername}`;
+    navigate(path);
```
Dennull: Being picky, can we inline the string interpolation on line 116 since we're not using the `path` variable anywhere else?

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+  };
+
+  const handleClear = () => {
+    setKeywords("");
+    setUserList([]);
+  };
+
+  const handleInputChange = (newInputValue: string) => {
+    debouncedSetter(newInputValue);
+    setFocusSearchBar(true);
```
Dennull: Does this need to be here? Doesn't the `onBlur` and `onFocus` handle it? Same on line 112

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+  autocomplete: {
+    "&.MuiAutocomplete-input": { paddingX: 0 },
+    position: "relative",
+  },
+  box: {
+    paddingBottom: 1,
+    paddingTop: 0,
+    paddingX: 2,
+  },
+  displayName: {
```
Dennull: Can we set the typography variant to `subtitle1` so that we can get rid of this?

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+    fontWeight: "bold",
+  },
+  listBox: {
+    maxHeight: "60vh",
+    overflowY: "auto",
+    zIndex: 1300,
```
Dennull: Is the overflow and the zIndex needed?

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+      color: "primary.main",
+    },
+  },
+};
+
 const SearchBar = ({ placeholder }: SearchBarProps) => {
+  const debouncedSetter = useMemo(
+    () => debounce((keywords: string) => setKeywords(keywords), 20),
+    []
+  );
```
Dennull: I think we want to debounce the `fetchUsers` request, not the setting of the keywords. Also maybe increase the debounce time to like 100 or 200 ms? 20 ms seems kinda short

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+  const [focusSearchBar, setFocusSearchBar] = useState(false);
+  const inputRef = useRef<HTMLInputElement>(null);
+  const [keywords, setKeywords] = useState("");
+  const [loading, setLoading] = useState(false);
+  const [userList, setUserList] = useState<SelectedUser[]>([]);
```
Dennull: Do you think `options` is a more clear name?

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
-          <InputAdornment position="start">
-            <SearchRoundedIcon />
-          </InputAdornment>
-        ),
-      }}
-      placeholder={placeholder}
-      size="small"
-    />
+    <Box sx={styles.box}>
+      <Autocomplete
```
Dennull: Maybe we should try setting the `freeSolo` prop to true? Then we can get rid of the clear logic and markup too since the freesolo mode already takes care of that

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+
+  const handleInputChange = (newInputValue: string) => {
+    debouncedSetter(newInputValue);
+    setFocusSearchBar(true);
+    if (newInputValue === "") {
+      setUserList([]);
+    }
+  };
+
+  const onSelect = (selectedUsername: string) => {
```
Dennull: We should set the `userList` to an empty array here so that when someone selects something the options array is empty

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+                      focusSearchBar
+                        ? styles.searchIconFocused
+                        : styles.searchIcon
+                    }
+                  >
+                    <SearchRoundedIcon />
+                  </IconButton>
+                </InputAdornment>
+              ),
+              type: "text",
```
Dennull: I believe the default input type is text, so we can delete this

src/components/Common/SearchBar.tsx
```diff
@@ -1,25 +1,211 @@
...
+              inputRef: inputRef,
+              onKeyDown: (e) => {
+                if (e.key === "Enter") {
+                  e.stopPropagation();
+                }
+              },
+            }}
```
Dennull: Having this stop propagation makes it so that you can't use the arrow keys to select an option from the dropdown menu, is that the behavior we'd like?

Dennull (changes requested):
I like this start, we can definitely build off of it. I feel like there's a lot to unpack with MUI's autocomplete component, so if you're free sometime and down to get on a call to go through the code let me know!

NasihNazeem:
Since there has been a few issues with rebasing, a change made from a linux machine that edited file line-endings, and the initial branching off of a feature branch — we have decided to pursue this feature on a separate branch and pull request (#126).

PR #109, opened 2024-07-03
T93, T112: Delete posts

### LINKED ISSUE #93: Deleting posts
When users click on the 3 dot icon on the top right of a post, a flyover should pop up that gives them the option to delete the post. This flyover should be easily customizable so that in the future we can add more additional post actions such as editing posts

There should also be checks in place to make sure that users can only delete their own posts. If a user clicks on the 3 dot icon and they don't own that post, nothing should happen for the sake of this ticket

### LINKED ISSUE #112: Post info menu
Each post item and expanded post item will have a kebab menu which will expand into a dropdown/menu. This menu will include functionality relative to the post such as being able to edit the post, delete the post or copy the link to the post.

It is important that only the author of the post should be able to edit or delete the post. Copying the link is public to anybody who has access to the post.

NasihNazeem:
## Related Issues

Closes #93 and closes #112

## Summary

This branch includes two features. In this branch we create a post menu for each post item and expanded post item. This menu allows users to edit, delete posts and copy link of post. It is crucial that the author of the post is the only one with access to edit and delete the post. Copy link is public to anybody who has access to the post

The second feature is integrating deletion of the post functionality

## Changes Made

To start, I created a new `deletePost` reducer in the store for posts. This allowed me to instantly make changes to the post list. I created a delete request to the backend that passes parameters such as the postId and the userId. To use on both the expanded post item and post items, I created a reusable component called PostMenu that encapsulates most of the functionality in this PR

## Screenshots

Open Menu

Open Menu

## Testing Instructions

The template is still up for debate so please do suggest any changes you think are appropriate

### DIFF
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -32,6 +32,7 @@ import { toggleLikePostRequest } from "../../utilities/postUtilities";
 import { Link as Routerlink } from "react-router-dom";
 import UserAvatar from "../Common/UserAvatar";
 import { useTheme } from "@mui/material/styles";
+import PostMenu from "./PostMenu";
 
 const styles = {
   actionButton: {
@@ -111,11 +112,7 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
       </Box>
       <CardHeader
         avatar={<UserAvatar username={post.username} />}
-        action={
-          <IconButton>
-            <MoreVertIcon />
-          </IconButton>
-        }
+        action={<PostMenu authorId={post.userId} postId={post.postId} />}
         title={
           <Link
             color={theme.typography.subtitle1.color}
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -117,7 +117,9 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
       </Box>
       <CardHeader
         avatar={<UserAvatar username={post.username} />}
-        action={<PostMenu authorId={post.userId} postId={post.postId} />}
+        action={
+          <PostMenu authorId={post.userId} postId={post.postId} isExpanded />
+        }
         title={
           <Link
             color={theme.typography.subtitle1.color}
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -2,23 +2,24 @@ import {
   Box,
   Button,
   Card,
+  CardActions,
   CardContent,
+  CardHeader,
+  CardMedia,
+  Divider,
+  IconButton,
+  Link,
   Stack,
   Typography,
-  Divider,
-  Link,
 } from "@mui/material";
-import KeyboardBackspaceIcon from "@mui/icons-material/KeyboardBackspace";
-import CardHeader from "@mui/material/CardHeader/CardHeader";
-import MoreVertIcon from "@mui/icons-material/MoreVert";
-import IconButton from "@mui/material/IconButton/IconButton";
-import CardMedia from "@mui/material/CardMedia/CardMedia";
-import CardActions from "@mui/material/CardActions/CardActions";
-import FavoriteBorderOutlinedIcon from "@mui/icons-material/FavoriteBorderOutlined";
-import FavoriteOutlinedIcon from "@mui/icons-material/FavoriteOutlined";
-import AddCommentOutlinedIcon from "@mui/icons-material/AddCommentOutlined";
-import RepeatOutlinedIcon from "@mui/icons-material/RepeatOutlined";
-import ShareOutlinedIcon from "@mui/icons-material/ShareOutlined";
+import {
+  AddCommentOutlined,
+  FavoriteBorderOutlined,
+  FavoriteOutlined,
+  KeyboardBackspace,
+  RepeatOutlined,
+  ShareOutlined,
+} from "@mui/icons-material";
 import axios from "axios";
 import { useAppDispatch, useAppSelector } from "../../state/hooks";
 import { useNavigate, useParams } from "react-router-dom";
@@ -88,16 +89,20 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
 
   useEffect(() => {
     const updatedExpandedPost = async () => {
-      const backupFetch = await axios.get(
-        "http://localhost:3001/api/posts/fetchPost",
-        {
-          params: {
-            userId: user.userId,
-            postId: urlParams.postId,
-          },
-        }
-      );
-      dispatch(setExpandedPost(backupFetch.data as Post));
+      try {
+        const backupFetch = await axios.get(
+          "http://localhost:3001/api/posts/fetchPost",
+          {
+            params: {
+              userId: user.userId,
+              postId: urlParams.postId,
+            },
+          }
+        );
+        dispatch(setExpandedPost(backupFetch.data as Post));
+      } catch (error) {
+        console.error("Failed to fetch post", error);
+      }
     };
     updatedExpandedPost();
   }, [dispatch, user.userId, urlParams.postId]);
@@ -106,7 +111,7 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
     <Card sx={styles.card}>
       <Box sx={styles.topHeader}>
         <IconButton onClick={() => navigate(-1)} sx={styles.backButton}>
-          <KeyboardBackspaceIcon color="secondary" />
+          <KeyboardBackspace color="secondary" />
         </IconButton>
         <Typography variant="h2">Post</Typography>
       </Box>
@@ -186,14 +191,14 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
           sx={styles.cardActions}
         >
           <IconButton>
-            <RepeatOutlinedIcon />
+            <RepeatOutlined />
           </IconButton>
           <IconButton
             onClick={() => {
               setOpen(true);
             }}
           >
-            <AddCommentOutlinedIcon />
+            <AddCommentOutlined />
           </IconButton>
           <IconButton
             onClick={() => {
@@ -207,13 +212,13 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
             sx={post.isLikedByCurrentUser ? styles.likedIcon : undefined}
           >
             {post.isLikedByCurrentUser ? (
-              <FavoriteOutlinedIcon />
+              <FavoriteOutlined />
             ) : (
-              <FavoriteBorderOutlinedIcon />
+              <FavoriteBorderOutlined />
             )}
           </IconButton>
           <IconButton>
-            <ShareOutlinedIcon />
+            <ShareOutlined />
           </IconButton>
         </Stack>
       </CardActions>
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -83,6 +83,7 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
   const user = useAppSelector((state) => state.user);
   const urlParams = useParams();
   const [open, setOpen] = useState(false);
+  const navigate = useNavigate();
 
   useEffect(() => {
     const updatedExpandedPost = async () => {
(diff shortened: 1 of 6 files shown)


src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -2,23 +2,24 @@ import {
...
-import CardHeader from "@mui/material/CardHeader/CardHeader";
-import MoreVertIcon from "@mui/icons-material/MoreVert";
-import IconButton from "@mui/material/IconButton/IconButton";
-import CardMedia from "@mui/material/CardMedia/CardMedia";
-import CardActions from "@mui/material/CardActions/CardActions";
-import FavoriteBorderOutlinedIcon from "@mui/icons-material/FavoriteBorderOutlined";
-import FavoriteOutlinedIcon from "@mui/icons-material/FavoriteOutlined";
-import AddCommentOutlinedIcon from "@mui/icons-material/AddCommentOutlined";
-import RepeatOutlinedIcon from "@mui/icons-material/RepeatOutlined";
-import ShareOutlinedIcon from "@mui/icons-material/ShareOutlined";
```
Dennull, line 21: I think specifying the route like this reduces our bundle size, so we might want to revert changes like this. I'll have to dig a little to find where exactly I heard that and if it's significant tho

src/components/Posts/PostItem.tsx
```diff
@@ -54,15 +56,24 @@ const styles = {
...
   },
+  menu: {
+    borderRadius: 4,
+  },
+  menuItem: {
+    display: "flex",
+    alignItems: "center",
+    justifyContent: "space-between",
+    width: "100%",
+  },
```
Dennull: Are these used?

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { useAppDispatch, useAppSelector } from "../../state/hooks";
+import { MoreVert, Edit, Delete, Link } from "@mui/icons-material";
+import axios from "axios";
+import { deletePost } from "../../state/slices/postsSlice";
+import { useNavigate } from "react-router-dom";
+
+type PostMenuProps = {
+  authorId: number;
+  postId: number;
+  isExpanded?: boolean;
```
Dennull: Nitpick, can we name this to `isExpandedPost`?

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+};
+
+const styles = {
+  listItemIcon: {
+    "&.MuiListItemIcon-root": { color: "black.main" },
+  },
```
Dennull: A small thing, instead of applying this style to the `ListItemIcon`, should we add it directly to the icons? That way we don't have to worry about the selector and can just say:
```suggestion
  icon: {
    color: "black.main",
  },
```

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+        `http://localhost:3001/api/posts/deletePost`,
+        {
+          data: {
+            postId: postId,
+            userId: userId,
+          },
+        }
+      );
+
+      isExpanded ? handleExpandedPostDelete() : handlePostDelete();
```
Dennull: My personal preference is to inline code whenever we can because it's easier to read than to jump around function definitions. Since this code isn't overly complex, can we put the code paths inside the function instead? I also don't think we need to return the data here since it got deleted/the state got updated in the store. Something like
```suggestion
      if(isExpanded) {
        dispatch(deletePost(postId));
      }
      else {
        navigate(-1)
      }
      setMenuOpen(false);
```

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+      return result.data;
+    } catch (error) {
+      console.error("Failed to delete the post", error);
+    } finally {
+      handleMenuClose();
+    }
+  };
+
+  const handleMenuOpen = (event: React.MouseEvent<HTMLElement>) => {
```
Dennull: Similar here, my preference is to inline this in the `onClick` below

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+    navigate(-1);
+  };
+
+  const handleTemporary = () => {
+    handleMenuClose();
+  };
+
+  return (
+    <>
+      <IconButton onClick={handleMenuOpen}>
```
Dennull: Can we make add the ref here instead of setting it on the `onClick`? The ref should always be on this icon button anyways since that's the element we want the menu to appear attached to, so I think we can save some code

If that's the case we don't need the state setter either, so we can use `useRef` instead

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <MoreVert />
+      </IconButton>
+      <Menu
```
Dennull, line 49: It sort of bothers me that on the hover the padding on the top and bottom of the list aren't highlighted.. I changed the color of the hover in the screenshot here so you can see:

image

My suggestion is we get rid of the padding of the menu list by setting the `MenuListProps` prop and then change the padding of the `MenuItem` instead. Maybe try something like making the right and left padding of the menu items 1.5 and the top and bottom padding 1. What do you think?

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+            <ListItemText
+              primaryTypographyProps={{
+                ...styles.listItemText,
+                color: "error",
+              }}
+            >
+              Delete Post
+            </ListItemText>
+          </MenuItem>,
+        ]}
```
Dennull: There's some weird syntax going on here with array stuff lol, can you change it back to normal? The entire thing after `userId === authorId` is wrapped in square brackets

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+      <Menu
+        anchorEl={anchorEl}
+        open={menuOpen}
+        onClose={handleMenuClose}
+        PaperProps={{
+          sx: styles.menu,
+        }}
+      >
+        {userId === authorId && [
+          <MenuItem onClick={handleTemporary} key="edit">
```
Dennull: I don't think we need keys for static items like this. I think react only needs keys when you're iterating through an array using a .map() because react has to know when the state changes for rerenders. But here, since all of the buttons are static, nothing will ever get rerendered, hence we can get rid of the key. See this [stackoverflow post](https://stackoverflow.com/questions/65702082/why-list-items-need-unique-key-but-other-not-in-react) that I read

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,130 @@
...
+            <ListItemIcon sx={styles.listItemIcon}>
+              <Edit />
+            </ListItemIcon>
+            <ListItemText primaryTypographyProps={styles.listItemText}>
```
Dennull: I suggest we use the typography variant `subtitle1` here. It's what we use for display names, and is bolded. I think it's the same font size too. Would be nice to cut down on the amount of custom fonts floating around

Dennull (changes requested):
Had a few suggestions, but the feature looks clean, I like it!!

Dennull:
If you're up for it, I'd love to see some extra improvements around confirmation too. Usually when you delete things on apps there's a confirmation dialog, so I think that'd be cool to add as an extra step right before the post deletes

We should also add a toast to display that the post was deleted, but I think there's a draft ticket for toasts in the backlog already

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: Okay, will look into this myself and see if it needs to be addressed

src/components/Posts/PostItem.tsx
NasihNazeem: Good catch sir

src/components/Posts/PostMenu.tsx
NasihNazeem: I modified it to inline code for the if statement, but I kept the menu closing inside of the finally block with a separate function definition since there are two functions I call within the handler and I use the handler in other elements as well

src/components/Posts/PostMenu.tsx
NasihNazeem: I feel like using inline here reduces the readability of the code, and if I keep a handler for opening the menu, we could use it for any confirmation modal

EDIT: I'm just reading up on your suggestion to use a ref instead, if we do that we might not need that handler funciton

src/components/Posts/PostMenu.tsx
NasihNazeem: Perfect, that works much better thank you

src/components/Posts/PostMenu.tsx
NasihNazeem: Since Menu doesn't like using fragments I tried using an array instead for the two items only show if its the current user's post. I will make it two separate conditional renders for the two items if we do not want to use the array

src/components/Posts/PostMenu.tsx
NasihNazeem: Thanks that clears it up, will try to avoid this

src/components/Posts/PostMenu.tsx
Dennull: Nice I like this better

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+  authorId: number;
+  postId: number;
+  isExpandedPost?: boolean;
+};
+
+const styles = {
+  icon: {
+    color: "black.main",
+  },
+  listItemText: { fontWeight: "bold" },
```
Dennull: We can get rid of this style now that we're using the typography variants

src/components/Posts/PostMenu.tsx
Dennull: Ohh I see, I didn't know that the menu doesn't let us use fragments. I'm ok with this, if we want to get rid of the multiple conditional renders we can wrap it in a `Box` too, but I'm fine either way

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <MoreVert />
+      </IconButton>
+      <Menu
+        anchorEl={menuRef.current}
+        open={menuOpen}
+        onClose={() => setMenuOpen(false)}
+        PaperProps={{
+          sx: styles.menu,
+        }}
+        MenuListProps={{ sx: { padding: 0 } }}
```
Dennull: Nit, can we put this in a styles object?

src/components/Posts/PostMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+};
+
+const PostMenu = ({
+  authorId,
+  postId,
+  isExpandedPost = false,
+}: PostMenuProps) => {
+  const userId = useAppSelector((state) => state.user.userId);
+  const menuRef = useRef<HTMLButtonElement>(null);
+  const [deleteModal, setDeleteModal] = useState(false);
```
Dennull: Can this be called `deleteConfirmationModalOpen`?

src/components/Posts/PostDeleteModal.tsx
Dennull, line 1: Do you think `PostDeleteConfirmationModal` is more descriptive? If you have time I'd also like to hop on a quick call to go over some minor styling decisions about the contents of the modal

Dennull (changes requested):

src/components/Posts/PostMenu.tsx
NasihNazeem: Removed

src/components/Posts/PostMenu.tsx
NasihNazeem: Updated the name

src/components/Posts/PostMenu.tsx
NasihNazeem: I took your feedback on this, how do you think it looks?

src/components/Posts/PostMenu.tsx
NasihNazeem: Updated!

src/components/Posts/PostDeleteModal.tsx
NasihNazeem: We renamed it to PostDeleteConfirmationModal

src/components/Posts/ExpandedPostItem.tsx
Dennull: As discussed offline, we can probably leave it the way you have it. Looking at MUI's docs, Vite probably has tree shaking built in, and either way, I don't think it's a big deal because our bundle size is probably fine anyways. This would be one of those optimizations we make when we actually need it

Dennull (approved):

PR #114, opened 2024-07-25
T106: Follower/Following list modal

### LINKED ISSUE #106: Follower/Following list
Implementing a modal that displays the list of users a profile is following and a list of users following the profile. Requirements:

- A modal pops up when the user clicks on either the following or followers tab on the viewed profile.
- UI/UX is similar to what Instagram has in their web application.
- Implement infinite scroll for the list of followers/following users.
- Ensure the follow/unfollow actions are updated in real time.

christopherkinyua:
Closes [#106 ](https://github.com/Project-Chirp/chirp-frontend/issues/106)

## Summary
This pull request introduces a modal that displays a list of users that the viewed profile follows and users that follow the account. This information is presented on the profile page when the user clicks either the link to see the followers or the following user list 

## Changes Made
I started by creating a custom FollowListModal which handles all the rendering of the list. The profile page uses the logic of acquiring the data using REST APIs of the endpoints I created in the backend. I pass the data to a useState array of type NetworkUsers that I have defined.

### DIFF
--- a/src/components/Common/FollowButton.tsx
+++ b/src/components/Common/FollowButton.tsx
@@ -14,7 +14,7 @@ const styles = {
 };
 
 type FollowButtonProps = {
-  onClick?: (e: React.MouseEvent) => void;
+  onClick?: () => void;
   visitedUserId: number;
 };
 
@@ -36,7 +36,7 @@ const FollowButton = ({ onClick, visitedUserId }: FollowButtonProps) => {
           },
         }
       );
-      onClick?.(e);
+      onClick?.();
     } catch (error) {
       console.log(error);
     }
--- a/src/components/Common/FollowButton.tsx
+++ b/src/components/Common/FollowButton.tsx
@@ -14,7 +14,7 @@ const styles = {
 };
 
 type FollowButtonProps = {
-  onClick?: () => void;
+  onClick?: (e: React.MouseEvent) => void;
   visitedUserId: number;
 };
 
@@ -36,7 +36,7 @@ const FollowButton = ({ onClick, visitedUserId }: FollowButtonProps) => {
           },
         }
       );
-      onClick?.();
+      onClick?.(e);
     } catch (error) {
       console.log(error);
     }
--- a/src/components/Common/FollowingButton.tsx
+++ b/src/components/Common/FollowingButton.tsx
@@ -1,7 +1,7 @@
 import { Button } from "@mui/material";
 import { useAppSelector } from "../../state/hooks";
 import axios from "axios";
-import { useState } from "react";
+import React, { useState } from "react";
 
 const styles = {
   followingButton: {
@@ -22,7 +22,7 @@ const styles = {
 };
 
 type FollowingButtonProps = {
-  onClick?: () => void;
+  onClick?: (e: React.MouseEvent) => void;
   visitedUserId: number;
 };
 
@@ -46,7 +46,7 @@ const FollowingButton = ({ onClick, visitedUserId }: FollowingButtonProps) => {
           },
         }
       );
-      onClick?.();
+      onClick?.(e);
     } catch (error) {
       console.log(error);
     }
--- a/src/components/Common/FollowingButton.tsx
+++ b/src/components/Common/FollowingButton.tsx
@@ -1,7 +1,7 @@
 import { Button } from "@mui/material";
 import { useAppSelector } from "../../state/hooks";
 import axios from "axios";
-import React, { useState } from "react";
+import { useState } from "react";
 
 const styles = {
   followingButton: {
--- a/src/components/Common/FollowingButton.tsx
+++ b/src/components/Common/FollowingButton.tsx
@@ -22,7 +22,7 @@ const styles = {
 };
 
 type FollowingButtonProps = {
-  onClick?: (e: React.MouseEvent) => void;
+  onClick?: () => void;
   visitedUserId: number;
 };
 
@@ -46,7 +46,7 @@ const FollowingButton = ({ onClick, visitedUserId }: FollowingButtonProps) => {
           },
         }
       );
-      onClick?.(e);
+      onClick?.();
     } catch (error) {
       console.log(error);
     }
--- a/src/components/Profile/FollowListModal.tsx
+++ b/src/components/Profile/FollowListModal.tsx
@@ -1,6 +1,5 @@
 import {
   Box,
-  Typography,
   Dialog,
   IconButton,
   DialogTitle,
@@ -12,10 +11,10 @@ import {
   ListItemAvatar,
   ListItemSecondaryAction,
   ListItemText,
+  Divider,
 } from "@mui/material";
 import CloseIcon from "@mui/icons-material/Close";
-import { User } from "@auth0/auth0-react";
-import { useState, useEffect } from "react";
+import { useState } from "react";
 import SearchBar from "../Common/SearchBar";
 import FollowingButton from "../Common/FollowingButton";
 import FollowButton from "../Common/FollowButton";
@@ -40,32 +39,30 @@ const styles = {
     display: "flex",
     alignItems: "center",
     justifyContent: "center",
-    padding: "8px 16px",
+    padding: 1,
     position: "relative",
   },
   closeIcon: {
     position: "absolute",
-    right: 16,
+    right: 6,
   },
   dialog: {
     borderRadius: 5,
+    boxSizing: "border-box",
   },
   searchBarContainer: {
-    padding: "8px 16px",
+    padding: 1,
   },
   listItem: {
     display: "flex",
     alignItems: "center",
-    padding: "8px 16px",
+    padding: 1,
   },
   avatar: {
-    marginRight: 16,
+    marginRight: 4,
   },
   listItemText: {
     flexGrow: 1,
-  },
-  divider: {
-    margin: "8px 0",
   },
 };
 
@@ -75,16 +72,14 @@ export default function FollowListModal({
   onClose,
 }: FollowListModalProps) {
   //Dummy Data
-  const [list, setList] = useState<User[]>([
+  const [list, setList] = useState<user[]>([
     {
-      id: "1",
       userName: "haileyhotrodhottie",
       displayName: "Hailey 🚗💄💋🦸‍♂️",
       imageURL: "https://via.placeholder.com/40",
       isFollowing: true,
     },
     {
-      id: "2",
       userName: "me_mo_ri",
       displayName: "Memori",
       imageURL: "https://via.placeholder.com/40",
@@ -93,10 +88,10 @@ export default function FollowListModal({
   ]);
 
   //Need to add proper logic for actual data
-  const handleFollowToggle = (id: string) => {
+  const handleFollowToggle = (index: number) => {
     setList((prevList) =>
-      prevList.map((user) =>
-        user.id === id ? { ...user, isFollowing: !user.isFollowing } : user
+      prevList.map((user, i) =>
+        i === index ? { ...user, isFollowing: !user.isFollowing } : user
       )
     );
   };
@@ -111,22 +106,20 @@ export default function FollowListModal({
     >
       <DialogTitle sx={styles.header}>
         {listType}
-        <IconButton onClick={onClose}>
+        <IconButton onClick={onClose} sx={styles.closeIcon}>
           <CloseIcon />
         </IconButton>
       </DialogTitle>
 
-      <Box sx={styles.divider}>
-        <hr />
-      </Box>
+      <Divider />
 
       <DialogContent>
         <Box sx={styles.searchBarContainer}>
           <SearchBar placeholder="Search Chirp" />
         </Box>
         <List>
-          {list.map((item) => (
-            <ListItem key={item.id} sx={styles.listItem}>
+          {list.map((item, index) => (
+            <ListItem key={index} sx={styles.listItem}>
               <ListItemAvatar>
                 <Avatar src={item.imageURL} sx={styles.avatar} />
               </ListItemAvatar>
@@ -138,13 +131,13 @@ export default function FollowListModal({
               <ListItemSecondaryAction>
                 {item.isFollowing ? (
                   <FollowingButton
-                    visitedUserId={parseInt(item.id)}
-                    onClick={() => handleFollowToggle(item.id)}
+                    visitedUserId={index}
+                    onClick={() => handleFollowToggle(index)}
                   />
                 ) : (
                   <FollowButton
-                    visitedUserId={parseInt(item.id)}
-                    onClick={() => handleFollowToggle(item.id)}
+                    visitedUserId={index}
+                    onClick={() => handleFollowToggle(index)}
                   />
                 )}
               </ListItemSecondaryAction>
--- a/src/components/Profile/FollowListModal.tsx
+++ b/src/components/Profile/FollowListModal.tsx
@@ -1,9 +1,16 @@
-import { Box, Typography, Dialog, IconButton } from "@mui/material";
+import {
+  Box,
+  Typography,
+  Dialog,
+  IconButton,
+  DialogTitle,
+  DialogContent,
+} from "@mui/material";
 import CloseIcon from "@mui/icons-material/Close";
 
 type FollowListModalProps = {
-  open: boolean;
-  handleClose: () => void;
+  openModal: boolean;
+  onClose: () => void;
 };
 
 const styles = {
@@ -18,24 +25,26 @@ const styles = {
   },
 };
 
-const FollowListModal = ({ open, handleClose }: FollowListModalProps) => {
+export default function FollowListModal({
+  openModal,
+  onClose,
+}: FollowListModalProps) {
   return (
     <Dialog
       fullWidth
-      onClose={handleClose}
-      open={open}
+      onClose={onClose}
+      open={openModal}
       scroll="paper"
       PaperProps={{ sx: styles.dialog }}
     >
-      <Box sx={styles.header}>
-        <Typography variant="h6">Followers</Typography>
-        <IconButton onClick={handleClose}>
+      <DialogTitle sx={styles.header}>
+        Followers
+        <IconButton onClick={onClose}>
           <CloseIcon />
         </IconButton>
-      </Box>
-      {/* Additional stuff goes here */}
+      </DialogTitle>
+
+      <DialogContent>{/* Additional stuff goes here */}Whats up</DialogContent>
     </Dialog>
   );
-};
-
-export default FollowListModal;
+}
--- a/src/components/Profile/FollowListModal.tsx
+++ b/src/components/Profile/FollowListModal.tsx
@@ -14,14 +14,17 @@ import {
   Divider,
 } from "@mui/material";
 import CloseIcon from "@mui/icons-material/Close";
-import { useState } from "react";
 import SearchBar from "../Common/SearchBar";
 import FollowingButton from "../Common/FollowingButton";
 import FollowButton from "../Common/FollowButton";
+import { useAppSelector } from "../../state/hooks";
+import { ProfileContent } from "../../pages/Profile";
+import React from "react";
+import { useNavigate } from "react-router-dom";
 
-export type user = {
+export type NetworkUsers = {
   userId: number;
-  userName: string;
+  username: string;
   displayName: string;
   imageURL: string;
   isFollowing: boolean;
@@ -30,9 +33,13 @@ export type user = {
 export type ListType = "Followers" | "Following";
 
 type FollowListModalProps = {
+  loading: boolean;
   openModal: boolean;
   listType: ListType | null;
-  listUserData: user[];
+  listUserData: NetworkUsers[];
+  profileContents: ProfileContent;
+  setProfileContents: (data: ProfileContent) => void;
+  setListUserData: (data: NetworkUsers[]) => void;
   onClose: () => void;
 };
 
@@ -50,18 +57,24 @@ const styles = {
   },
   dialog: {
     borderRadius: 5,
-    boxSizing: "border-box",
+    width: "25%",
+    height: "50%",
   },
   searchBarContainer: {
-    padding: 1,
+    paddingX: 2,
+    paddingY: 1,
   },
   listItem: {
     display: "flex",
     alignItems: "center",
-    padding: 1,
+    paddingX: 2,
+    "&:hover": {
+      backgroundColor: "gray.light",
+      cursor: "pointer",
+    },
   },
   avatar: {
-    marginRight: 4,
+    marginRight: 2,
   },
   listItemText: {
     flexGrow: 1,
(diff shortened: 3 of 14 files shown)


src/pages/Profile.tsx
```diff
@@ -124,6 +128,35 @@ const Profile = () => {
...
+
+      const result = await axios.get(endpoint, {
+        params: {
+          visitedUserId: profileContents.userId,
+          currentUserId: currentUserId,
+        },
+      });
+      setListData(result.data);
+    } catch (error) {
+      console.error(error);
```
Dennull: Let's not throw an error, maybe just console log it or something. Ideally we'd handle it gracefully by showing the user an error message but I think there's a ticket for error handling. We shouldn't throw an error though or else the user is forced to restart the app

src/pages/Profile.tsx
```diff
@@ -124,6 +128,35 @@ const Profile = () => {
   });
   const [editProfileModalOpen, setEditProfileModalOpen] = useState(false);
   const [loading, setLoading] = useState(true);
+  const [networkModalLoading, setNetworkModalLoading] = useState(true);
```
Dennull: Can we please name this like `followListModalLoading` or something? The component is named follow list but this is named network modal so it's a little inconsistent

src/pages/Profile.tsx
```diff
@@ -124,6 +128,35 @@ const Profile = () => {
...
+  const [openNetworkModal, setOpenNetworkModal] = useState(false);
```
Dennull: Same here maybe go with `followListModelOpen` or something

src/pages/Profile.tsx
```diff
@@ -124,6 +128,35 @@ const Profile = () => {
...
+  const [listType, setListType] = useState<ListType | null>(null);
+  const [listData, setListData] = useState<NetworkUsers[]>([]);
```
Dennull: Similar here, can you please make the name a bit more descriptive? Like `followerListData`

src/pages/Profile.tsx
```diff
@@ -124,6 +128,35 @@ const Profile = () => {
...
+  const handleOpenModal = async (listType: ListType) => {
+    setListType(listType);
+    setOpenNetworkModal(true);
+
+    try {
+      setNetworkModalLoading(true);
+      const endpoint =
+        listType === "Followers"
+          ? "http://localhost:3001/api/follow/getFollowersUserList"
+          : "http://localhost:3001/api/follow/getFollowingUserList";
```
Dennull: Ngl I'm not a huge fan of this `listType` variable, and especially it being a string. What do you think about just making 2 separate requests whenever the buttons are clicked? So we'd basically handle all of this in the `onClick` of each button separately instead of trying to combine both of them

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+  profileContents,
+  setProfileContents,
+  setListUserData,
+  onClose,
+}: FollowListModalProps) => {
+  const currentUserId = useAppSelector((state) => state.user.userId);
+  const navigate = useNavigate();
+
+  const handleFollowToggle = async (index: number, e: React.MouseEvent) => {
+    e.stopPropagation();
```
Dennull: I think both `FollowButton` and `FollowingButton` already do this, so we might be able to chuck it and get rid of the event parameter too. Now that I think about it the way we're doing those buttons might not be ideal, so they could use a refactor, but that's out of the scope of this ticket

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      const isFollowing: boolean = updatedList[index].isFollowing;
+      const newFollowingCount = isFollowing
+        ? ++profileContents.followingCount
+        : --profileContents.followingCount;
+
+      setProfileContents({
+        ...profileContents,
+        followingCount: newFollowingCount,
+      });
+    }
```
Dennull: It feels like this should be done in the `Profile` component and this should be passed as a callback function. Maybe the callback can be called something like `onFollowed`?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+  };
+
+  const handleNavigation = (username: string) => {
+    onClose();
+    navigate(`/${username}`);
+  };
```
Dennull: Can we please inline this in the `onClick`?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  const modifiedListUserData = listUserData.sort((a, b) =>
```
Dennull: This is cool, can we rename it to something including the word "sorted" rather than "modified"?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      <DialogTitle sx={styles.header}>
+        {listType}
+        <IconButton onClick={onClose} sx={styles.closeIcon}>
+          <CloseIcon />
+        </IconButton>
+      </DialogTitle>
+
+      <Divider />
+      <Box sx={styles.searchBarContainer}>
+        <SearchBar placeholder="Search Chirp" />
```
Dennull: Nitpick, can this placeholder be just "Search"?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      </Box>
+
+      <DialogContent sx={{ paddingBottom: 1 }}>
```
Dennull: Should put this in a styles object

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const navigate = useNavigate();
+
+  const handleFollowToggle = async (index: number, e: React.MouseEvent) => {
+    e.stopPropagation();
+    const updatedList = [...listUserData];
+    updatedList[index] = {
+      ...listUserData[index],
+      isFollowing: !listUserData[index].isFollowing,
+    };
+    setListUserData(updatedList);
```
Dennull: This works, but usually in this case we map through the list, find the item with the id we want, then change that one item. Can we follow that paradigm for consistency? I think it reads a bit better as well

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      <Box sx={styles.searchBarContainer}>
+        <SearchBar placeholder="Search Chirp" />
+      </Box>
+
+      <DialogContent sx={{ paddingBottom: 1 }}>
+        {!loading && (
+          <List>
+            {modifiedListUserData.map((item, index) => (
+              <ListItem
+                key={index}
```
Dennull: Nit, can we make the key the user's id?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+      <Divider />
+      <Box sx={styles.searchBarContainer}>
+        <SearchBar placeholder="Search Chirp" />
+      </Box>
+
+      <DialogContent sx={{ paddingBottom: 1 }}>
+        {!loading && (
+          <List>
+            {modifiedListUserData.map((item, index) => (
```
Dennull: Minor nitpick, can we use `o` to refer to the current iteration in the map? It makes it simpler so unless there's a specific reason not to (like nested maps, or maybe cases where you're accessing both the current iteration and the entire list) it cuts down on the amount of bloat. This is kinda just a coding convention we've been following

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+  setListUserData: (data: NetworkUsers[]) => void;
+  onClose: () => void;
+};
+
+const styles = {
+  header: {
+    display: "flex",
+    alignItems: "center",
+    justifyContent: "center",
+    padding: 1,
```
Dennull: We usually have the "X" to close the modal on the left of the modal. Look at `EditProfileModal` for an example of the styling or let me know if you want help

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+    position: "relative",
```
Dennull: I don't think this does anything

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+  },
+  closeIcon: {
+    position: "absolute",
+    right: 6,
+  },
```
Dennull: We should use `position: "absolute"` sparingly because it breaks the document flow. Again see `EditProfileModal` for an example of this

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      backgroundColor: "gray.light",
+      cursor: "pointer",
+    },
+  },
+  avatar: {
+    marginRight: 2,
+  },
+  listItemText: {
+    flexGrow: 1,
+  },
```
Dennull: What does this do?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+                onClick={() => handleNavigation(item.username)}
+              >
+                <ListItemAvatar>
+                  <Avatar src={item.imageURL} sx={styles.avatar} />
+                </ListItemAvatar>
+                <ListItemText
+                  primary={item.displayName}
+                  secondary={item.username}
+                  sx={styles.listItemText}
+                  primaryTypographyProps={{ fontWeight: "bold" }}
```
Dennull: Can this go into the styles object?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      <Divider />
+      <Box sx={styles.searchBarContainer}>
+        <SearchBar placeholder="Search Chirp" />
+      </Box>
+
+      <DialogContent sx={{ paddingBottom: 1 }}>
+        {!loading && (
+          <List>
+            {modifiedListUserData.map((item, index) => (
+              <ListItem
```
Dennull: We haven't had many of these list items that have 2 actions.. I think we should follow how MUI does theirs on their [docs](https://mui.com/material-ui/react-list/#checkbox). They nest a  `ListItemButton` inside of a `ListItem`. 

This would also takes care of the styling so we can get rid of the hover styling you've added to the list item

src/pages/Profile.tsx
```diff
@@ -257,12 +290,14 @@ const Profile = () => {
                     </Box>
                   )}
                 </Box>
+
                 <Box sx={styles.followerContainer}>
                   <Link
                     color={theme.palette.black.main}
                     component={Routerlink}
                     to={`/${username}`} // TODO: Create Modal to check followers
```
Dennull: Remove this comment and the one on line 310 too please

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,189 @@
...
+      PaperProps={{ sx: styles.dialog }}
+    >
+      <DialogTitle sx={styles.header}>
+        {listType}
+        <IconButton onClick={onClose} sx={styles.closeIcon}>
+          <CloseIcon />
+        </IconButton>
+      </DialogTitle>
+
+      <Divider />
```
Dennull, line 97: I think adding some more padding between the divider, the header and the search bar would look better

Dennull (changes requested):

src/pages/Profile.tsx
christopherkinyua: Done

src/pages/Profile.tsx
christopherkinyua: Done, done, done.

src/pages/Profile.tsx
christopherkinyua: Done

src/pages/Profile.tsx
christopherkinyua: Removed outdated comments

src/components/Profile/FollowListModal.tsx
christopherkinyua: Yeah I see it. Doesnt seem to do anything. Thanks for the catch

src/pages/Profile.tsx
christopherkinyua: So do I just basically change the `console.error` to `console.log(error)` ?

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done

src/components/Profile/FollowListModal.tsx
christopherkinyua: Sure, done

src/components/Profile/FollowListModal.tsx
christopherkinyua: Added some padding. Let me know if it looks good to you

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done

src/pages/Profile.tsx
christopherkinyua: Done as stated above

src/components/Profile/FollowListModal.tsx
christopherkinyua: Okay, done. So basically any time I am mapping an array I should name it `o` for coding conventions.

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done, remove this

src/components/Profile/FollowListModal.tsx
christopherkinyua: Changed it as per the implementation in `EditProfileModal`

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done

src/pages/Profile.tsx
christopherkinyua: Okay, I basically made two separate functions for each list type. I hope that improves readability.

src/components/Profile/FollowListModal.tsx
christopherkinyua: Like remove the `e.stopPropagation`? Yeah I see that it is done in the Follow and Following Buttons. I will remove that line.

src/components/Profile/FollowListModal.tsx
christopherkinyua: Sure, I've done this. Kinda makes sense so I'm not doing too much.

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done, thanks for the help

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done with your help. Thanks

src/pages/Profile.tsx
Dennull: Yeah you got it

Dennull:
As discussed offline, we'll convert this back to a draft while we figure out the best way to move forward. I think there's some confusion between how we're managing client side and server side state. If we want to introduce the react query library into the app we might want to change how we're managing our server side state in general, so we should probably fully understand that before implementing infinite scroll

src/components/Common/FollowButton.tsx
```diff
@@ -14,7 +14,7 @@ const styles = {
 };
 
 type FollowButtonProps = {
-  onClick?: () => void;
+  onClick?: (e: React.MouseEvent) => void;
```
Dennull: We don't need the `e` parameter anymore right? Same for `FollowingButton`?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+export type NetworkUsers = {
+  userId: number;
+  username: string;
+  displayName: string;
+  imageURL: string;
+  isFollowing: boolean;
+};
+
+export type ListType = "Followers" | "Following";
```
Dennull: I'm not a fan of string types like this, I think it should be an enum. I think we can also get away without needing to define this tho, see some of my other comments

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+} from "@mui/material";
+import CloseIcon from "@mui/icons-material/Close";
+import SearchBar from "../Common/SearchBar";
+import FollowingButton from "../Common/FollowingButton";
+import FollowButton from "../Common/FollowButton";
+import { useAppSelector } from "../../state/hooks";
+import React from "react";
+import { useNavigate } from "react-router-dom";
+
+export type NetworkUsers = {
```
Dennull, line 26: I don't really like this name tbh lol but I think I'll make a separate ticket for consolidating some of our user types

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+  displayName: string;
+  imageURL: string;
+  isFollowing: boolean;
+};
+
+export type ListType = "Followers" | "Following";
+
+type FollowListModalProps = {
+  loading: boolean;
+  openModal: boolean;
```
Dennull: Can this just be `modalOpen` or just `open`?

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+    const updatedList = listUserData.map((o) => {
+      if (userId === o.userId) {
+        onFollowed(o.isFollowing);
+        return { ...o, isFollowing: !o.isFollowing };
+      }
+      return o;
+    });
+
+    setListUserData(updatedList);
+  };
```
Dennull: It feels like this should be pushed up one level to the parent component. My intent was to have all of this run in the `onFollowed` callback function. So basically move all of this logic to the parent and pass this function into `onFollowed`, then call `onFollowed` after someone clicks the follow button

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+        </Box>
+      </DialogTitle>
+
+      <Divider />
+      <Box sx={styles.searchBarContainer}>
+        <SearchBar placeholder="Search" />
+      </Box>
+
+      <DialogContent sx={styles.dialogContent}>
+        {!loading && (
```
Dennull, line 103: Should we display a loader is `loading` is true?

src/pages/Profile.tsx
```diff
@@ -124,6 +128,59 @@ const Profile = () => {
...
+  const [followerListData, setFollowerListData] = useState<NetworkUsers[]>([]);
+
+  const onFollowed = (isFollowing: boolean) => {
+    setProfileContents((prevProfileContents) => ({
+      ...prevProfileContents,
+      followingCount: isFollowing
+        ? --prevProfileContents.followingCount
+        : ++prevProfileContents.followingCount,
+    }));
+  };
```
Dennull: We can probably handle the `followerListData` updates here too rather than inside the `FollowListModal`. That way can get rid passing the `setFollowerListData` setter function

src/pages/Profile.tsx
```diff
@@ -331,6 +390,17 @@ const Profile = () => {
           open={editProfileModalOpen}
         />
       )}
+      {!followListModalLoading && followListModelOpen && (
```
Dennull: We probably don't need the `!followListModalLoading` condition here. We can either delay setting `followListModalOpen` until the modal until the data has been fetched or show the loader inside of the modal. Either way, we don't need that extra conditional render

src/pages/Profile.tsx
```diff
@@ -124,6 +128,59 @@ const Profile = () => {
   });
   const [editProfileModalOpen, setEditProfileModalOpen] = useState(false);
   const [loading, setLoading] = useState(true);
+  const [followListModalLoading, setFollowListModalLoading] = useState(true);
+  const [followListModelOpen, setFollowListModelOpen] = useState(false);
```
Dennull: Slight typo here, it should be `followListModal`

src/components/Profile/FollowListModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+          <List>
+            {sortedListUserData.map((o) => (
+              <ListItem key={o.userId} sx={styles.listItem}>
+                <ListItemButton
```
Dennull, line 107: I took a look at this and I'm unsure about having nested buttons. In general it isn't best practice to nest interactive elements like buttons or links. I think Twitter does it anyways because they don't care, and we agree with that is up for debate, but my vote is to for now change this to be non-nested buttons until we get to talk about it. I'll make a PR with my suggested styling changes into this one

src/pages/Profile.tsx
```diff
@@ -124,6 +128,59 @@ const Profile = () => {
...
+    setProfileContents((prevProfileContents) => ({
+      ...prevProfileContents,
+      followingCount: isFollowing
+        ? --prevProfileContents.followingCount
+        : ++prevProfileContents.followingCount,
+    }));
+  };
+
+  const handleOpenFollowersModal = async () => {
+    setListType("Followers");
```
Dennull: I'm not really a fan of this `ListType` thing we have going on here. Firstly, I don't like that it's a string, I think if we wanted to do types like this they should be enums

I'd prefer if we get rid of it entirely though, cause something about passing flags into components to tell them where to fetch data from seems wrong to me. Passing in the url endpoint also feels greasy too. My opinion is that the lowest level components shouldn't have to worry about the data; all they need to know is the _shape_ of the data being passed to them, and then they can render the UI appropriately

So for our case, I'm proposing we have 2 separate modals in here and conditionally render them based on what the user clicked. Since the endpoints to grab the follower and following data are completely different, we should fetch for them separately in the calling component and just pass them into a generic follower/following modal component. Then as for the modal title, we can make that into a prop

Dennull (changes requested):

src/components/Profile/FollowListModal.tsx
Dennull: Created #123 for this

src/components/Profile/FollowListModal.tsx
christopherkinyua: Changed it to `open`

src/pages/Profile.tsx
christopherkinyua: Makes sense, I just removed the `followListModalLoading` condition as in the async calls, it is set false after it has received the data. Not sure if there is anything else I have to do regarding this

src/pages/Profile.tsx
christopherkinyua: Done, Ive updated the code to have two separate modals that are conditionally rendered and I changed the `listType` prop to be `title` instead.

src/pages/Profile.tsx
christopherkinyua: Done

src/components/Profile/FollowListModal.tsx
christopherkinyua: Removed based on the other comments given

src/components/Profile/FollowListModal.tsx
christopherkinyua: Sure thing. Ill mark this as resolved

src/components/Common/FollowButton.tsx
christopherkinyua: Done

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done

src/components/Profile/FollowListModal.tsx
christopherkinyua: Done, I moved the `onFollowed` logic into this function and passed it as a callback function within `FollowListModal`

src/pages/Profile.tsx
christopherkinyua: Done

src/components/Common/FollowingButton.tsx
```diff
@@ -1,7 +1,7 @@
 import { Button } from "@mui/material";
 import { useAppSelector } from "../../state/hooks";
 import axios from "axios";
-import { useState } from "react";
+import React, { useState } from "react";
```
Dennull: We don't need this right?

src/components/Profile/FollowListModal.tsx
Dennull: Where's the loader?

src/pages/Profile.tsx
```diff
@@ -124,6 +126,68 @@ const Profile = () => {
   });
   const [editProfileModalOpen, setEditProfileModalOpen] = useState(false);
   const [loading, setLoading] = useState(true);
+  const [followListModalLoading, setFollowListModalLoading] = useState(true);
+  const [followListModalOpen, setFollowListModalOpen] = useState(false);
+  const [followListModalUserData, setFollowListModalUserData] = useState<
+    NetworkUsers[]
+  >([]);
+  const [isFollowersModal, setIsFollowersModal] = useState(false);
+  const [isFollowingModal, setIsFollowingModal] = useState(false);
```
Dennull: I don't think we need both of these along with `followListModalOpen`. We can make a `followerListModalOpen` and `followingListModalOpen` to cut it down to 2 states

Dennull (changes requested):

src/components/Common/FollowingButton.tsx
christopherkinyua: True

src/pages/Profile.tsx
christopherkinyua: Sure thing.

Dennull (approved):

PR #119, opened 2024-09-04
T115: Edit posts

### LINKED ISSUE #115: Edit posts
We are creating a modal to handle the editing of the posts.

- This will be present in the timeline and the profile page

Implementation
- Will make a modal that resembles the post modal; pre-load the text component with the details of the original post and include an update button
- Create an endpoint and SQL query that deals with the updating of the post

christopherkinyua:
## Related Issues

Closes #115

## Summary

This PR aims to introduce the feature of editing a post within the application

## Changes Made

The changes made to realize this was as follows:
- Created a modal that will handle the UI for editing a post
- Updated the post slice to handle the state update.
- Modified `formatTimestamp` function to handle the `editedTimestamp` value

## Screenshots

image

## Testing Instructions

No additional notes


src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -159,7 +162,9 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
       )}
       <Box sx={styles.timestampBox}>
         <Typography component="span" variant="subtitle2">
-          {formatTimestamp(post.timestamp)}
+          {post.editedTimestamp
+            ? formatTimestamp(post.editedTimestamp, true)
+            : formatTimestamp(post.timestamp)}
```
Dennull: I kind of don't like how we have to do this ternary everywhere. We might be able to simplify this a bit by using Javascript's [boolean coercion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean)

Instead of passing in a plain boolean, we can either use the `Boolean()` function or the double not operator `!!` to cast the edited timestamp to a boolean. If we do this, the edited timestamp will evaluate to `false` if it's `undefined` or an empty string `""`, and `true` otherwise

Similarly, we can use [short circuit evaluation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_OR#short-circuit_evaluation) by using the `||` operator to pass the right timestamp into the function. We'd make it so that by default the edited timestamp gets passed in, and if that's not present we'll pass the original timestamp:

```suggestion
         {formatTimestamp(
            post.editedTimestamp || post.timestamp,
            Boolean(post.editedTimestamp)
          )}
```

@NasihNazeem @christopherkinyua let me know if you guys find this suggestion better. At the end of the day it's just about which one is easier to read, so I'm happy to go with whichever one

Dennull: Another alternative would be to change the function arguments of `formatTimestamp` to do something similar. If instead of taking in `isEdited` as a boolean we change the parameters to be something like `timestamp` and `editedTimestamp`, we can do a similar check within the function: if  the edited timestamp is present, use that and go with the edited formatting, else use the normal timestamp and format it as usual

src/components/Posts/PostMenu.tsx
```diff
@@ -8,12 +8,16 @@ import {
...
+import EditPostModal from "./EditPostModal";
 import PostDeleteConfirmationModal from "./PostDeleteConfirmationModal";
 
 type PostMenuProps = {
   authorId: number;
   postId: number;
+  postTextContent: string;
+  postTimeStamp: string;
   isExpandedPost?: boolean;
+  prevEditedPostTimestamp: string;
```
Dennull: Do you think we can drop the post prefix on these things? I think it's obvious enough that these props refer to that one post. So it'd change to `textContent`, `timestamp` and `editedTimestamp` if so

src/components/Posts/PostMenu.tsx
```diff
@@ -33,12 +37,16 @@ const styles = {
...
+  postTextContent,
+  postTimeStamp,
   isExpandedPost = false,
+  prevEditedPostTimestamp,
 }: PostMenuProps) => {
   const userId = useAppSelector((state) => state.user.userId);
   const menuRef = useRef<HTMLButtonElement>(null);
   const [deleteConfirmationModalOpen, setDeleteConfirmationModalOpen] =
     useState(false);
+  const [editModal, setEditModal] = useState(false);
```
Dennull: Can this be something like `editModalOpen`?

src/components/Posts/PostMenu.tsx
```diff
@@ -56,7 +64,7 @@ const PostMenu = ({
         MenuListProps={{ sx: styles.menuList }}
       >
         {userId === authorId && (
-          <MenuItem sx={styles.menuItem} onClick={() => setMenuOpen(false)}>
+          <MenuItem sx={styles.menuItem} onClick={() => setEditModal(true)}>
```
Dennull: Right now, if you click the edit post button, the edit post modal opens and once you're done with everything you have to click again to close the post menu. Do we also want to run `setMenuOpen(false)` on this onClick so that the post menu closes on click, or are we ok with the current behavior?

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+import EmojiPickerIconButton from "../Common/EmojiPickerIconButton";
+import UserAvatar from "../Common/UserAvatar";
+import AddPhotoAlternateOutlinedIcon from "@mui/icons-material/AddPhotoAlternateOutlined";
+import formatTimestamp from "../../utilities/formatTimestamp";
+import axios from "axios";
+import { updatePost } from "../../state/slices/postsSlice";
+
+type EditPostModalProps = {
+  onClose: () => void;
+  openModal: boolean;
```
Dennull: Can this just be `open`?

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+  postId: number;
+  originalPostTextContent: string;
+  originalPostTimeStamp: string;
+  prevEditedPostTimestamp: string;
```
Dennull: Nit, the camel casing is off here. It should be `originalPostTimestamp` and `prevEditedPostTimestamp`. A suggestion too, can we shorten the name to `editedPostTimestamp`?

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+  openModal,
+  postId,
+  originalPostTextContent,
+  originalPostTimeStamp,
+  prevEditedPostTimestamp,
+}: EditPostModalProps) => {
+  const [postTextContent, setPostTextContent] = useState("");
+  const user = useAppSelector((state) => state.user);
+  const dispatch = useAppDispatch();
+  const existingPost = useAppSelector((state) =>
```
Dennull, line 64: If we're looking up the existing post from the store already, we don't need to pass the original post text content and timestamp through the props. We can grab all of those from this object

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+    state.posts.posts.find((post) => post.postId === postId)
+  );
+
+  useEffect(() => {
+    setPostTextContent(originalPostTextContent);
+  }, [originalPostTextContent]);
```
Dennull: We shouldn't need this useEffect, we can set the initial state of `postTextContent` instead. For example, right now it'd be something like:

```
const [postTextContent, setPostTextContent] = useState(originalPostTextContent);
```

But we can take it a step further and just extract that directly from the store object so we can ditch the props too:

```
const existingPost = useAppSelector((state) =>
  state.posts.posts.find((post) => post.postId === postId)
);
const [postTextContent, setPostTextContent] = useState(
  existingPost?.textContent
);
```

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+        {
+          postId: postId,
+          textContent: postTextContent,
+          editedTimestamp: new Date(),
+        },
+        {
+          headers: {
+            "Content-Type": "application/json",
+          },
+        }
```
Dennull: I actually don't think we need this header stuff because axios takes care of it by default

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+                      (prevContent) => prevContent + emoji.emoji
+                    );
+                  }}
+                />
+              </Stack>
+              <Button
+                size="small"
+                type="submit"
+                variant="contained"
+                onClick={handleEdit}
```
Dennull: For consistency let's wrap the entire thing in a `form` and make this into a submit button. I don't think it makes a huge difference but since we're using a from in our other dialogs let's follow the same pattern here. The button actually already has `type="submit"` attribute on it, probably from a copy paste somewhere lol

If you do this you'll remove the `onClick` and assign `handleEdit` to the `onSubmit` of the form instead

src/components/Posts/EditPostModal.tsx
Dennull, line 1: This dialog looks slightly different from the other dialogs that users use to create posts.. For example, I don't know if we need the username, display name, and timestamp. I'd like to meet and go over what we want our compose/edit post modals to look like

I think this can be for another ticket though, because we have a few modals floating around now and I'd also like to see if we can standardize the look of them (for example, padding, how the title looks, how the action buttons look)

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,190 @@
...
+          editedTimestamp: new Date(),
+        },
+        {
+          headers: {
+            "Content-Type": "application/json",
+          },
+        }
+      );
+
+      if (existingPost) {
```
Dennull: The `existingPost` object is typed as `string | undefined` because of the `.find()` operation, but in reality we know it should never be undefined. The post id doesn't come from a user input, it comes from our code, so we should always get a post back

Either way, I think we can handle this situation more gracefully than this. I'd propose we either use a type assertion using the non-null assertion operator (`!`) or throw an error inside here. @NasihNazeem do you have any thoughts?

Dennull (changes requested):

src/components/Posts/PostMenu.tsx
christopherkinyua: Done

src/components/Posts/PostMenu.tsx
christopherkinyua: Done

src/components/Posts/PostMenu.tsx
christopherkinyua: Yeah, i think adding `setMenuOpen(false);` is a good idea. It will also keep consistency with how the deletePost one works.

src/components/Posts/EditPostModal.tsx
christopherkinyua: Done

src/components/Posts/EditPostModal.tsx
christopherkinyua: Makes sense. Done

src/components/Posts/EditPostModal.tsx
christopherkinyua: Alright. Removed

src/components/Posts/EditPostModal.tsx
christopherkinyua: Bruh, this makes sense, I was wondering how to re-render after I have defined the the `useState` to handle the `originalTextContent`. I will be on the look out for instances I can use the already defined states.

christopherkinyua: Done

src/components/Posts/EditPostModal.tsx
christopherkinyua: I was looking at our figma designs earlier today and I saw we had decided to implement our dialogs as the way we did with `ComposePost`. I will move forward with that direction. I will start with the refactoring of `ComposePost` so it can be used instead of the `editPostModal`

src/components/Posts/EditPostModal.tsx
christopherkinyua: done

src/components/Posts/EditPostModal.tsx
christopherkinyua: Ive implemented both just to be safe. Can be overkill but cant be too safe

src/components/Posts/ExpandedPostItem.tsx
christopherkinyua: Yoo, this is some cracked thinking. I didn't know about short circuit evaluation and boolean coercion. Ill keep this in mind for future implementations.

Done

src/components/Posts/EditPostModal.tsx
christopherkinyua: When I implement this, I am getting an error: `Uncaught TypeError: Cannot read properties of undefined (reading 'textContent')`  when I implement the non-null assertion as follows:

`const [postTextContent, setPostTextContent] = useState<string>(
existingPost!.textContent
   );`
   
From my understanding, it is implying that at one point it is `undefined` which I can inference from what you said above about the `.find()` operation. I have tried a work around using `useEffect` to update the render if the `existingPost` changes which it does since we know it can never be null and then update the `useState` for the `postTextContent`.

Let me know if this is graceful enough. I am just trying stuff to make sure it works and I am not getting errors.

src/components/Common/TooltipTimestamp.tsx
```diff
@@ -33,7 +35,9 @@ const TooltipTimestamp = ({
       title={formatTooltipTimestamp(timestamp)}
     >
       <Typography variant={variant} sx={styles.tooltipText}>
-        {formatTimestamp(timestamp)}
+        {isEdited
+          ? formatTimestamp(timestamp, true)
+          : formatTimestamp(timestamp)}
```
Dennull: This can be simplified:
```suggestion
        {formatTimestamp(timestamp, isEdited)}
```

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+  },
+};
+
+const EditPostModal = ({ onClose, open, postId }: EditPostModalProps) => {
+  const user = useAppSelector((state) => state.user);
+  const dispatch = useAppDispatch();
+  const existingPost = useAppSelector((state) =>
+    state.posts.posts.find((post) => post.postId === postId)
+  );
+  const [postTextContent, setPostTextContent] = useState<string>("");
```
Dennull: This is just so you know, but you need to put the non-null assertion at the end for this to work. Adding the useEffect below adds more code for no reason. Also, we probably don't need to type the useState here:

```suggestion
  const [postTextContent, setPostTextContent] = useState(existingPost.textContent!);
```

If we go with my suggestion above to apply the non-null assertion on the `exisingPost` object itself we won't need to do any of this anyways

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+    marginTop: 1,
+  },
+};
+
+const EditPostModal = ({ onClose, open, postId }: EditPostModalProps) => {
+  const user = useAppSelector((state) => state.user);
+  const dispatch = useAppDispatch();
+  const existingPost = useAppSelector((state) =>
+    state.posts.posts.find((post) => post.postId === postId)
+  );
```
Dennull: Let's put the non-null assertion up here so we don't have to propagate it everywhere. This shouldn't ever really be undefined because there should always be a post with that postId. Maybe we can handle the error better if it were to ever happen but that can be an exercise to do some other time

```suggestion
  const existingPost = useAppSelector((state) =>
    state.posts.posts.find((post) => post.postId === postId)
  )!;
```

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+      setPostTextContent(existingPost.textContent);
+    }
+  }, [existingPost]);
+
+  const onSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    if (!existingPost) {
+      throw new Error(`Post with ID ${postId} not found`);
+    }
```
Dennull: Adding this error and also doing the non-null assertion below is redundant; by adding this check you're telling typescript that anywhere below this point `existingPost` is never undefined, so the non-null assertion does nothing

If move the non-null assertion to line 64 where `existingPost` is define we can do away with this error and the assertion on line 89 anyways tho

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+  );
+  const [postTextContent, setPostTextContent] = useState<string>("");
+
+  useEffect(() => {
+    if (existingPost) {
+      setPostTextContent(existingPost.textContent);
+    }
+  }, [existingPost]);
+
+  const onSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
```
Dennull: Everywhere else we use `React.SyntheticEvent` for the event parameter, let's stick to that for consistency. I actually don't really know the difference lol

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -158,7 +159,12 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
         />
       )}
       <Box sx={styles.timestampBox}>
-        <TooltipTimestamp timestamp={post.timestamp} />
+        {
```
Dennull: Do we need the braces?

src/components/Posts/PostMenu.tsx
```diff
@@ -101,6 +109,11 @@ const PostMenu = ({
         postId={postId}
         isExpandedPost={isExpandedPost}
       />
+      <EditPostModal
+        onClose={() => setEditModalOpen(false)}
+        open={editModalOpen}
+        postId={postId}
+      ></EditPostModal>
```
Dennull: Let's self close this

```suggestion
      <EditPostModal
        onClose={() => setEditModalOpen(false)}
        open={editModalOpen}
        postId={postId}
      />
```

src/components/Posts/EditPostModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+      await axios.put("http://localhost:3001/api/posts/editPost", {
+        postId: postId,
+        textContent: postTextContent,
+      });
+      dispatch(
+        updatePost({
+          ...existingPost!,
+          textContent: postTextContent,
+        })
+      );
```
Dennull: You also need to update the `editedTimestamp` here

Dennull (changes requested):

src/components/Common/TooltipTimestamp.tsx
christopherkinyua: Done

src/components/Posts/EditPostModal.tsx
christopherkinyua: Yeah, I was doing too much. I removed it from the code.

src/components/Posts/EditPostModal.tsx
christopherkinyua: Sure thing. Done.

src/components/Posts/ExpandedPostItem.tsx
christopherkinyua: Nop, I removed it

src/components/Posts/PostMenu.tsx
christopherkinyua: Done

src/components/Posts/EditPostModal.tsx
christopherkinyua: Makes sense. Poor coding on my part.

src/components/Posts/EditPostModal.tsx
christopherkinyua: Done

src/components/Posts/EditPostModal.tsx
christopherkinyua: Done

src/components/Posts/EditPostModal.tsx
Dennull: No, like I meant on the front end. I've done it for you though

Dennull (approved):

src/components/Posts/EditPostModal.tsx
christopherkinyua: Ahh, I didnt think i had to use the `Tostring()` for the date

PR #121, opened 2024-09-05
T120: Display full timestamp on hover

### LINKED ISSUE #120: Show full timestamp on hover
In this ticket, we want to display the full date of when of the post timestamp when we hover over it

christopherkinyua:
## Related Issues

Closes #120 

## Summary

This PR is to address having the timestamp display the full time and date of when the post was made

## Changes Made

To achieve this I:
- Utilized the ToolTip component from MUI
- Modified `formatTimestamp` function to have an `isToolTip` boolean which returns the full date and time

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


Dennull:
Can we also add this feature to our messages?

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -158,9 +159,15 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
...
-        </Typography>
+        <Tooltip
+          title={formatTimestamp(post.timestamp, true)}
+          placement="bottom"
+          arrow
+        >
+          <Typography component="span" variant="subtitle2">
+            {formatTimestamp(post.timestamp)}
+          </Typography>
+        </Tooltip>
```
Dennull: I think it might be a good idea to make a component for this, for example `TooltipTimestamp`. We might want to use this in other places and keep the styling the same

I don't like how far away the tooltip is from the timestamp tho, can we change the `slotProps` to make the distance less? See MUI's docs: https://mui.com/material-ui/react-tooltip/#distance-from-anchor

Dennull: Additionally, can we add a hover effect to the tooltip? Like change the font styling to underline and make the cursor a pointer when the user hovers?

src/components/Posts/PostItem.tsx
```diff
@@ -99,7 +103,17 @@ const PostItem = ({ post }: PostProps) => {
             </Link>
           </Box>
         }
-        subheader={formatTimestamp(post.timestamp)}
+        subheader={
+          <Box sx={styles.toolTipContainer}>
```
Dennull: If we add this container styling directly to `subheaderTypographyProps`, we can get rid of this `Box`

src/utilities/formatTimestamp.ts
```diff
@@ -1,8 +1,27 @@
-const formatTimestamp = (entry: string) => {
+const formatTimestamp = (entry: string, isToolTip?: boolean) => {
```
Dennull: It feels like this function is sort of starting to do a lot. In #119 the function already has a parameter that determines if it's an edited post, so adding another flag seems to bloat it more. Maybe we should make a new function called `formatTooltipTimestamp` to take care of this case?

@NasihNazeem and @christopherkinyua what do we think? Should we make a new function or are y'all ok with how this one would look when merged with the changes in the edit posts PR?

Dennull (changes requested):

src/utilities/formatTimestamp.ts
christopherkinyua: Yeah, making another function should be fine. I can make it. Having it as another function will make reading the code easier too.

christopherkinyua: Done

src/components/Posts/ExpandedPostItem.tsx
christopherkinyua: Yeah, making a separate component is a lot cleaner and a better way to do it. I've done that plus adding the hover effect

src/components/Posts/PostItem.tsx
christopherkinyua: Done

christopherkinyua:
@Dennull The PR is ready for your review again

src/pages/DirectMessage.tsx
```diff
@@ -226,9 +227,9 @@ const DirectMessage = () => {
                       </Box>
                     }
                     secondary={
-                      <Typography sx={styles.timestamp} variant="body2">
-                        {formatTimestamp(o.timestamp)}
-                      </Typography>
+                      <Box>
```
Dennull: Does this `Box` need to be here?

src/utilities/ToolTipTimestamp.tsx
Dennull, line 1: This file is a component, so it should be moved into `components/Common`

Dennull, line 1: I also made a few minor changes with their props that I pushed:

- Added the `arrow` prop back in
- I made it non-interactive by adding the `disableInteractive` prop. Now you can't hover over the tooltip itself, which I think is better for our use case
- Changed the offset from [5, -15] to [0, -5]

Let me know if you agree with these changes

src/utilities/ToolTipTimestamp.tsx
```diff
@@ -0,0 +1,35 @@
+import { Tooltip, Typography } from "@mui/material";
+import formatTimestamp from "./formatTimestamp";
+import formatToolTipTimestamp from "./formatToolTipTimestamp";
+
+type ToolTipTimestampProps = {
+  ToolTipTimestamp: string;
```
Dennull: Props should be camel case (ex. `tooltipTimestamp`). Also, can we change this name to something like `timestamp`?

src/utilities/ToolTipTimestamp.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Tooltip
+      title={formatToolTipTimestamp(ToolTipTimestamp)}
+      placement="bottom"
+      slotProps={{
+        popper: {
+          modifiers: [{ name: "offset", options: { offset: [5, -15] } }],
+        },
+      }}
+    >
+      <Typography variant="subtitle2" sx={styles.toolTipText}>
```
Dennull: Let's add a prop for `variant` and default it to `subtitle2`. In some cases we want to change this type, for example in `DirectMessage` it should be the `body2` variant. We can get the variant type from MUI's `TypographyVariant` type

So it'd look something like this:

```
type ToolTipTimestampProps = {
  ToolTipTimestamp: string;
  variant?: TypographyVariant;
};

...

const ToolTipTimestamp = ({ ToolTipTimestamp, variant = "subtitle2" }: ToolTipTimestampProps) => {
  return (
    <Tooltip
      placement="bottom"
      ...
    >
      <Typography variant={variant} sx={styles.toolTipText}>
        {formatTimestamp(ToolTipTimestamp)}
      </Typography>
    </Tooltip>
  );
};

```

Dennull (changes requested):

src/pages/DirectMessage.tsx
christopherkinyua: Nope, It was needed when I was initially working on it. Fixed it by removing it

src/utilities/ToolTipTimestamp.tsx
christopherkinyua: Moved to components

christopherkinyua: Thats nice, I had removed the arrow coz I was trying to emulate how X has their timestamp tooltip set up which is a lot closer to the text hence the arrow looked weird.

Otherwise, it looks good. I was struggling with the interactiveness of the tooltip so know of the `disableInteractive` is helpful. I approve these changes

src/utilities/ToolTipTimestamp.tsx
christopherkinyua: Done

src/utilities/ToolTipTimestamp.tsx
christopherkinyua: Thats a neat way to make it more dynamic. I have included the changes

christopherkinyua:
@Dennull PR is ready for your review again

Dennull (approved):

PR #125, opened 2024-09-18
T95, T124: Sharing posts, adding toasts

### LINKED ISSUE #95: Sharing posts
When users click on the share post button, the post link should be copied the clipboard. After, a toast notification should pop up to inform them that the post link was copied to their clipboard

### LINKED ISSUE #124: Toasts
A toast or snackbar is a small notification that pops up to give users feedback after an action. See https://mui.com/material-ui/react-snackbar/ for examples from MUI's `Snackbar` component. We'd like to integrate toasts for some actions, such as posting and deleting a post

Some requirements for the toast component are:
- The toast should be a global toast that can be used across the app, ideally called using a custom hook
- Expose MUI's `Snackbar` props in the hook so we can customize them
  - The toast position should be taken as a prop, but should have a default position of the top middle (see [MUI's docs](https://mui.com/material-ui/react-snackbar/#position) for more info about how to position their toast component)
  - The toast variant should also be a prop. For example, we can have a success variant, an error variant, ect (again, see [MUI's docs](https://mui.com/material-ui/react-snackbar/#use-with-alerts) for a good example of how we can use their `Alert` component to help with this)
  - The hide duration is another example of a prop it would be good to expose. The default should be 5 seconds
  - Bonus points if we can add additional actions using the hook (eg. an undo button, link, etc)
- Only 1 toast should be shown at a time. If the user clicks on actions that trigger multiple toasts, dismiss the old ones and only display the new one

Dennull:
## Related Issues

Closes #95 and #124

## Summary

Now when users click the share post button on either the post actions or the post menu, the post URL gets copied to their clipboard

Additionally, toasts were added to the app to give the user feedback on various actions such as copying links to their clipboard and deleting posts. Toasts can be queued by using the new `enqueueToast` action. If more than 1 toast is queued at a time, any previous toasts will be immediately dequeued until the latest toast is shown

## Changes Made

For sharing posts, we simply used the `navigator.clipboard.writeText()` function

For toasts, a new slice in the redux store was added along with a global `Toast` component. This makes it so that we don't have to manually render toasts everywhere; a single toast component in the app is shown or hidden. To queue a toast, simply dispatch the `enqueueToast` action. There are also few other actions, which the global `Toast` component uses to manage the toast state and queue under the hood

## Screenshots

image

image

## Testing Instructions

No additional notes


src/components/Posts/PostMenu.tsx
```diff
@@ -86,7 +92,16 @@ const PostMenu = ({
             </ListItemText>
           </MenuItem>
         )}
-        <MenuItem sx={styles.menuItem} onClick={() => setMenuOpen(false)}>
+        <MenuItem
```
NasihNazeem: What do you think about creating a separate function for the `onClick` on this MenuItem? Since there seems to be a few actions occuring, I suggest we move it away to make the jsx clearer

src/components/Posts/PostDeleteConfirmationModal.tsx
```diff
@@ -70,8 +71,20 @@ const PostDeleteModal = ({
       } else {
         dispatch(deletePost(postId));
       }
+      dispatch(
+        enqueueToast({
+          message: "Your post has been deleted",
+          severity: "info",
```
NasihNazeem: I feel like this is more of a success severity since its after an action that requests the deletion of a post.

Info severity can be used where we can provide a bit more context about the action or something like that

src/state/slices/toastSlice.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const initialState: ToastLoafState = {
+  currentToast: null,
+  loaf: [],
+  open: false,
+};
+
+const defaultToastOptions: Toast = {
+  anchorOrigin: { vertical: "top", horizontal: "center" },
+  autoHideDuration: 5000,
```
NasihNazeem: 5000 feels a bit too long, how about 3000?

src/components/Posts/PostDeleteConfirmationModal.tsx
```diff
@@ -70,8 +71,20 @@ const PostDeleteModal = ({
...
+        })
+      );
     } catch (error) {
       console.error("Failed to delete the post", error);
+      dispatch(
+        enqueueToast({
```
NasihNazeem, line 82: Toasts are naturally quite dismissive so I'm not sure how I feel about errors as toasts. If we want to maintain its usage for errors, I recommend creating a longer notification for errors compared to success, info or warning severities

src/components/Common/Toast.tsx
```diff
@@ -0,0 +1,60 @@
...
+      dispatch(dequeueToast());
+      dispatch(setToastOpen(true));
+    } else if (loaf.length > 0 && currentToast && open) {
+      // Close the current toast when a new one is added
+      dispatch(setToastOpen(false));
+    }
+  }, [currentToast, loaf, dispatch]);
+
+  const handleClose = (
+    event?: React.SyntheticEvent | Event,
```
NasihNazeem: Since this is an unused argument should we change this to `_?`

src/App.tsx
```diff
@@ -53,29 +54,32 @@ function App() {
...
-        path="/post/:postId"
-        element={<ProtectedRoute component={ExpandedPost} />}
-      />
-      <Route
-        path="/coming-soon"
-        element={<ProtectedRoute component={ComingSoon} />}
-      />
-    </Routes>
+    <>
+      <Toast />
```
NasihNazeem, line 58: What if we create a Provider component for Toast and use it within `index.tsx`? I feel like that will make it clearer as we're only returning the protected routes within App.tsx and we're still making toasts globally accessible.

Let me know what you think, because you can also argue that this is just creating too many components (Toast, and ToastProvider)

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -107,6 +108,10 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
     updatedExpandedPost();
   }, [dispatch, user.userId, urlParams.postId]);
 
+  const copyToClipboard = () => {
+    navigator.clipboard.writeText(`http://localhost:3000/post/${post.postId}`);
```
NasihNazeem, line 113: This isn't a suggestion but I wonder whether this function would work on mobile? If I remember correctly this is only used in certain browser APIs

NasihNazeem (changes requested):
Great work Dennis and Aninda. I had some suggestions and questions about how we want to use this but I can't wait to merge this into main.

src/components/Common/Toast.tsx
christopherkinyua: I was gonna ask since it isnt being used, what was the vision for it? I am trying to understand this whole toast thing

src/components/Posts/ExpandedPostItem.tsx
christopherkinyua: I checked the docs and it seems to support Mobile versions of web browsers. Correct me if I am wrong but I think `navigator.clipboard.writeText()` works with most modern web browser versions.

image

christopherkinyua:
Besides Nasih's comments, I can't think of anything else. I just need to understand the `setToastOpen` reducer in the ToastSlice file. This is some good work!

src/components/Posts/PostMenu.tsx
Dennull: Ok sure, I think I'll move it all into `copyToClipboard()`

src/state/slices/toastSlice.ts
Dennull: Sure, fine with me!

src/components/Posts/PostDeleteConfirmationModal.tsx
Dennull: I don't have a strong opinion either way, so I can remove it if we want. I can increase the `autoHideDuration` or change the message if we want, do you have any suggestions?

I have seen places where toasts are used for errors too though, for example when the connection is poor. I think any user feedback is better than the nothing we have right now, but we're kinda ignoring errors as it is anyways so that point might be moot. Again I'm down for anything, just lmk what y'all think is best

src/components/Common/Toast.tsx
Dennull: @christopherkinyua this is an event argument that gets supplied from MUI's callback, so we don't have control over the function signature. That event argument is going to be there whether we use it or not, but since we need access to the second `reason` argument they give us we need to include it in the function arguments

@NasihNazeem I remember I was having some typescript issues before which is why I kept this here, but for some reason it's not complaining anymore lol so I've changed it to `_`

src/components/Posts/ExpandedPostItem.tsx
Dennull: Thanks for looking into it @christopherkinyua. On the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText#browser_compatibility) it also says that most of these browser versions have been out for a few years now (around 2018) so we should be good for this

src/App.tsx
Dennull: I guess we can, but since we're using Redux we technically already have a provider since Redux is managing our toast state management. I can move the `Toast` component into `index.tsx` if we want, but the actual state management is taken care of already

NasihNazeem: Ah, I see what you mean. In that case I think this is fine, it doesn't affect the functionality

src/components/Posts/PostDeleteConfirmationModal.tsx
NasihNazeem: You're right in saying any feedback is better than nothing. Let's stick to it, I suggest increasing the `autoHideDuration` for error messages. Maybe `5000` works here like you had originally?

src/components/Posts/PostItem.tsx
```diff
@@ -156,7 +160,16 @@ const PostItem = ({ post }: PostProps) => {
           >
             {post.numberOfLikes}
           </Button>
-          <IconButton>
+          <IconButton
+            onClick={() => {
+              copyToClipboard();
+              dispatch(
```
NasihNazeem: I noticed you moved the dispatch calls in `PostMenu` and `ExpandedPostItem` inside of `copyToClipboard()` and I think we should do the same here just for consistency

NasihNazeem (changes requested):
Looks good to me, I replied to some of the previous conversation and also had this minor change for consistency. It's ready to be approved aside from that

src/App.tsx
Dennull: Yeah, technically if you wanted to show the toast outside of anything rendered in `App.tsx`, it wouldn't show. So you can change the _state_ of the toast in `index.tsx`, but nothing would show. It's probably fine to keep it here for now though since all of our current app functionality is in `App.tsx`

src/components/Posts/PostDeleteConfirmationModal.tsx
Dennull: Sounds good, I made the default `autoHideDuration` 5000 if severity is error in `toastSlice.ts` so that we don't have to remember to manually set it everywhere

NasihNazeem (approved):

PR #126, opened 2024-09-28
T85: Search bar

### LINKED ISSUE #85: App search bar dropdown

NasihNazeem:
## Related Issues

Closes #85 

## Summary

_**This is a repost of the initial search bar pull request**_: #107 

This pull request introduces a fully functional search bar that allows users to search for other users within the application by username or display name. The search results are ordered by the number of followers, from most to least. Each result displays a temporary avatar, the username, and the display name of the user

This enhancement aims to improve user experience by making it easier and faster to find other users within the app

## Changes Made

- Debouncing: Improves search performance by limiting the rate of search queries
- Clear search: Provides a clear button to reset the search input
- Blur on select: Automatically blurs the search input when a user is selected from the results

## Screenshots

image

## Testing Instructions

No additional notes


src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+import { useAuth0 } from "@auth0/auth0-react";
+import { SelectedUser } from "../../state/slices/messagesSlice";
+import { useNavigate } from "react-router-dom";
+
+type SearchBarDropDownProps = {
+  placeholder: string;
+};
+
+const styles = {
+  autocomplete: {
```
Dennull: I don't think we need this class

src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+    "&.MuiAutocomplete-hasClearIcon .MuiOutlinedInput-root": { paddingX: 1 },
+    position: "relative",
+    padding: 0,
+  },
+  box: {
```
Dennull: A suggestion, what do you think about removing this class? The search bar appears wider, and I could see us going with either

With `box` class:

image

Without `box` class:

image

src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+  useEffect(() => {
+    debouncedFetch(keywords);
+  }, [keywords]);
+
+  const handleInputChange = (newInputValue: string) => {
+    setKeywords(newInputValue);
+
+    if (newInputValue.trim() === "") {
+      setSearchOptions([]);
+    }
```
Dennull: We're handling this twice, should we leave the setting of the search options to `fetchUsers`? That way this function's only job is to handle what's displayed in the input field. We can also move this inline in that case

src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  return (
+    <Box sx={styles.box}>
+      <Autocomplete
+        disablePortal
+        fullWidth
+        filterOptions={(x) => x}
+        getOptionLabel={() => ""}
+        forcePopupIcon={false}
+        id="search"
```
Dennull: Unless we have a use for it, I think we can get rid of this

src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const debouncedFetch = useMemo(
+    () => debounce((keywords: string) => fetchUsers(keywords), 100),
+    []
+  );
+
+  const fetchUsers = async (keywords: string) => {
+    setLoading(true);
+    if (keywords.length === 0) {
+      return;
+    }
```
Dennull: In this code path `loading` never gets set back to false. Also, can we combine with the `keywords.trim() === ""` code path on line 89? Feels like they're doing the same thing

src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+        fullWidth
+        filterOptions={(x) => x}
+        getOptionLabel={() => ""}
+        forcePopupIcon={false}
+        id="search"
+        value={null}
+        inputValue={keywords}
+        ListboxProps={{ style: styles.listBox }}
+        loading={!keywords && loading}
+        loadingText="Start typing to search..."
```
Dennull: This is a minor thing, but it looks like we're mixing a few things up here. I think the `loading` prop shouldn't depend on `!keywords`. Also, the loading text should be something like "Loading..." (which is the default), since it should show when the fetch is occurring (ie. the user has already typed something in)

The other side of this is when the user hasn't typed anything in, in which case we should be showing "Start typing to search...", or when there are no results found, in which case we should show "No user found".

So, I feel like we should change these lines to handle only loading stuff and move the no options logic into `noOptionsText` like so:

```
noOptionsText={!keywords ? "Start typing to search...": "No user found"}
```

I tried it out an ran into another complication where the keywords gets set before `fetchUsers` runs, so this makes our `noOptionsText` flash for a split second before the loading state is set. So, we may have to set the loading state true in `handleInputChange` or something like that

src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+        onChange={(_, value) => value && handleSelect(value.username)}
+        onBlur={handleClear}
```
Dennull: Do we want the input field to clear on blur? Do you think it'd be better if we kept the input field and the options as is and close the dropdown on blur instead? I don't really have a strong opinion on this, so I'd be happy with either one. Let me know your thoughts!

src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+        onInputChange={(_, newInputValue) => handleInputChange(newInputValue)}
+        onFocus={() => setFocusSearchBar(true)}
+        open={focusSearchBar}
+        openOnFocus
```
Dennull: I think we only need one of these. If we're using `openOnFocus`, we shouldn't need to set the `open` prop, and vice versa

src/components/Common/SearchBarDropdown.tsx
```diff
@@ -0,0 +1,207 @@
...
+                        : styles.searchIcon
+                    }
+                  >
+                    <SearchRoundedIcon />
+                  </IconButton>
+                </InputAdornment>
+              ),
+              endAdornment: keywords && (
+                <InputAdornment position="end">
+                  <IconButton onClick={() => setKeywords("")}>
```
Dennull: The search options don't get cleared here, so when you click the clear button the dropdown options are still available

src/components/SideBar/SideBar.tsx
```diff
@@ -20,7 +20,7 @@ const SideBar = () => {
   return (
     <Box sx={styles.rightContent}>
       <Box sx={styles.searchBarContainer}>
-        <SearchBar placeholder="Search Chirp" />
+        <SearchBarDropdown placeholder="Search Chirp" />
```
Dennull: Nit, can we make the placeholder "Search users"? Same with the `ExpandedPostSidebar`

Dennull (changes requested):

src/components/Common/SearchBarDropdown.tsx
NasihNazeem: I like the second, it aligns well with the rest of the Sidebar components

src/components/Common/SearchBarDropdown.tsx
NasihNazeem: Sure thing that makes sense, I'll inline the setState for keywords

src/components/Common/SearchBarDropdown.tsx
NasihNazeem: Alright so I've moved the logic for `keywords.trim() ==== ""` to the if statement before the try catch block, and also setting loading to false inside of that.

src/components/Common/SearchBarDropdown.tsx
NasihNazeem: Cool so if I follow this method, there is a case of the loading state showing up for half a second or so due to loading with every keyword. I could void the loading state from the autocomplete in general because we do take care of it within the debounced fetchUsers call

src/components/Common/SearchBarDropdown.tsx
NasihNazeem: I guess from a user standpoint it would be kinda helpful if you are half-way through a search and you want to quickly look at something else. I can remove the clearing for onBlur by setting the `clearOnBlur` prop to false.

src/components/Common/SearchBarDropdown.tsx
NasihNazeem: Good point, I'll just use `handleClear` here instead where I clear the keywords and search options state

Dennull (approved):

PR #128, opened 2024-09-28
T67: Custom Axios hook

### LINKED ISSUE #67: Pass JWT token on all requests
All requests on the frontend need to pass the token to authorize the request and retrieve/update any data in the backend. The backend will have middleware running on all endpoints that verify the token that's been sent on any request.

Create a custom hook for all requests within the application and pass url, method and optional data parameters in a config object.

### LINKED ISSUE #127: Unfollow through following modal decrement incorrectly
When attempting to unfollow a user through the FollowingModal, the following users value on the profile bio seems to decrease by 2 even though it should be by one.

NasihNazeem:
## Related Issues

Closes #67

Linked Issue: #127

## Summary

This pull request introduces a custom `useAxios` hook to streamline API requests in our application. The hook integrates Auth0 for authentication and provides a unified interface for making authenticated HTTP requests.

Key features:

- Automatic token handling with Auth0
- Support for all the common HTTP methods (GET, POST, PUT, DELETE)
- Built-in loading and error states

This aims to simplify API calls, reduce the amount of boilerplate code and ensure consistent authentication and error handling.

## Changes Made

1. Created a new custom hook `useAxios`:
     - Implemented `useCallback` for memoized function calls (`sendRequest`)
     - Uses Auth0's `getAccessTokenSilently` for JWT verification
     - Uses axios for HTTP requests
     - Supports custom headers, request body and URL parameters
2. Implemented environment variable check for base url:
     - Created an env variable for the base_url including /api
     - Throws an error if the base url is not defined

Future Improvements:
- Error Handling could be improved depending on type of error
- Caching?
- Retries for certain errors
- Cancellation requests
- Interceptors

## Screenshots

No additional screenshots

## Testing Instructions

- The `VITE_BASE_URL` environment variable must be set for the hook to function correctly
- Consider if we want to add additional features like request cancellation or automatic retries in future iterations


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "ESNext",
     "lib": ["dom", "dom.iterable", "esnext"],
-    "types": ["vite/client", "vite-plugin-svgr/client"],
+    "types": ["vite/client", "vite-plugin-svgr/client", "node"],
```
Dennull: Do we need this?

package-lock.json
Dennull, line 1: This still has changes, are you sure you reverted it to match what's on main?

src/utilities/useAxios.ts
```diff
@@ -0,0 +1,54 @@
+import { useAuth0 } from "@auth0/auth0-react";
+import axios from "axios";
+import { useCallback } from "react";
+
+export type SendRequestProps = {
+  endpoint: string;
+  method?: "GET" | "POST" | "PUT" | "DELETE";
+  headers?: { [key: string]: string };
+  body?: any;
+  params?: any;
```
Dennull: There's a `AxiosRequestConfig` that we can use for this. Maybe we can simplify this to be a parameter called config or something similar and make it of type `AxiosRequestConfig`. Then we can spread that into the axios request inside of `sendRequest`. That way we have access to all of the axios config options too

src/utilities/useAxios.ts
```diff
@@ -0,0 +1,54 @@
...
+};
+
+const useAxios = () => {
```
Dennull, line 5: Instead of making the parameters something we send to the function itself, we can make them the parameters of `useAxios` and then pass them in. Do you think this is better? 

I'd also like to make the function parameters individual parameters instead of one big object that needs to be destructured. I think along with making a unified config parameter with the `AxiosRequestConfig` type this will make the function signature look nicer. For example

```
const useAxios = (endpoint: string, config: AxiosRequestConfig) =>
// Hook
```

src/utilities/useAxios.ts
```diff
@@ -0,0 +1,54 @@
...
+  const { getAccessTokenSilently } = useAuth0();
+
+  const sendRequest = useCallback(
```
Dennull, line 8: Is there a particular reason for using `useCallback`? Does the function run a lot without it?

src/utilities/useAxios.ts
Dennull, line 1: Axios lets us set some of the defaults for our config if we're interested in that. We can either set [them directly](https://axios-http.com/docs/config_defaults) (likely somewhere in a high level component such as `App.tsx` or `index.tsx`) or make our own [custom axios instance](https://axios-http.com/docs/instance) (probably in here)

This might be nice for us since all of our endpoints right now share a lot of configurations. However, if we ever need to call an external API or change our configuration for one of our endpoints we'd need a way to override that. I just thought this was a cool thing they offered, let me know what you think about it. I'm totally cool with looking more into this in a separate ticket

Dennull, line 1: I see you removed the loading and error states from earlier versions of the hook. I'm also in favor of making a new ticket to add those features in later. Just to confirm though, you weren't using those in any components to manage loading states right? They were all handling their loading states themselves? I took a glance and didn't find any that were using it but wanted to double check

src/components/Messages/ConversationList.tsx
```diff
@@ -67,7 +64,14 @@ const ConversationList = () => {
...
             conversation={o}
             onClick={() => {
-              navigate(`/messages/${user.userId}/${o.otherUserId}`);
+              dispatch(
+                setSelectedConversation({
+                  displayName: o.displayName,
+                  username: o.username,
+                  userId: o.otherUserId,
+                })
+              );
```
Dennull: I think you removed this in another PR, probably accidently brought it back in a merge?

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
```diff
@@ -1,32 +1,30 @@
...
-          params: {
-            userId: user.userId,
-          },
-        }
-      );
-      setConversationList(result.data as SelectedUser[]);
+      const result = await sendRequest({
+        endpoint: "messages/getModalConversations",
+        method: "GET",
+        params: { userId: userId },
```
Dennull: Nit, this can be shorted to just `params: { userId }` because the value name is the same as the key

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -87,28 +87,23 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
...
+        params: { userId: user.userId, postId: urlParams.postId },
+      });
+      dispatch(setExpandedPost(backupFetch as Post));
     };
     updatedExpandedPost();
   }, [dispatch, user.userId, urlParams.postId]);
 
   const copyToClipboard = () => {
+    console.log(window.location.origin);
+
```
Dennull: Console log here

src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -210,6 +205,7 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
           <IconButton
             onClick={() => {
```
Dennull: In `PostItem.tsx` there's an async await for `toggleLikePostRequest`, do we need that here too?

src/pages/Profile.tsx
```diff
@@ -144,23 +145,24 @@ const Profile = () => {
     setProfileContents((prevProfileContents) => ({
       ...prevProfileContents,
       followingCount: isFollowing
-        ? --prevProfileContents.followingCount
-        : ++prevProfileContents.followingCount,
+        ? prevProfileContents.followingCount - 1
+        : prevProfileContents.followingCount + 1,
```
Dennull: I don't think this actually fixes the bug. If you unfollow someone, refresh the page, then follow them again from the modal it appends to the string (ie. "4" turns to "41" instead of "5"). I debugged it and the reason why we're having so much trouble with this stuff is because the back end returns a string instead of a number, so we're trying to increment/decrement a string

I don't have an answer for question of whether the back end should return numbers as strings in the first place, but the temporary fix for this is to cast it to a number first

```suggestion
        ? Number(prevProfileContents.followingCount) - 1
        : Number(prevProfileContents.followingCount) + 1,
```

Dennull:
Nice job, this is much needed and will save quite a few lines of code. I like the suggestions you made about enhancing this feature more with cancellation requests and intercepts; I agree that there are a lot more enhancements we can make around our requests. For example, I think cancellation would be a good thing to add because we can prevent some memory leaks when components unmount before a request finishes

Another thing that might be good to look into is whether we should use a `useEffect` in our custom request hook, since a lot of components need to fetch data as a side effect anyways. This would prevent the us from needing to write a `useEffect` in almost every component manually. Then there's the whole thing about moving error and loading states from the components to the request hook too

However, we have talked about possibly migrating to another library that might abstract that all away for us like react query. We also currently don't do any of these extra things now anyways, so maybe we explore all of that in a separate ticket. My thought is let's just get this thing merged to take care of the auth tokens since we're about to secure all of the endpoints on the back end. The hook will make it easier to modify all of our requests in one spot down the line after it's merged too. Anyways, all this to say we'll make new tickets to track our improvements around this hook

Dennull (changes requested):

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
NasihNazeem: good catch

src/components/Messages/ConversationList.tsx
NasihNazeem: you right my bad

package-lock.json
NasihNazeem: Ok I think I reverted it now hopefully it changes. I tried doing it earlier but maybe an `npm i` is making those changes? I'm not sure if that's how it works

src/utilities/useAxios.ts
NasihNazeem: yeah that's a good idea instead of constricting us to only using these 4 objects

src/utilities/useAxios.ts
NasihNazeem: Yes I like the idea of having single parameters such as endpoint and config. I'd like to keep the parameters within the function itself as passing them with the hook will exclusively work on single endpoint components and we have a few components with multiple endpoints. We could initialize the hook multiple times, but I am not a fan of that solution as it feels like unneccessary duplication

src/utilities/useAxios.ts
NasihNazeem: I kinda kept this here as a precaution because I think it is good practice to have a useCallback on a function that runs in a useEffect. It could lead to unneccessary re-renders, so I was thinking we can memoize the function to avoid that

src/utilities/useAxios.ts
NasihNazeem: I was looking into this too, all hooks have the same auth header and the content-type isn't exactly dynamic. Same with our base url. I'm willing to make this change but if it is not too much trouble, I think having a separate ticket to make the defaults would be better

NasihNazeem: Yes you're right, I did not use it in any as far as I can tell. I reviewed to be sure, but I tried to make it so that if there was a component with a loading state already, I wouldn't import the loading state from the hook. If I did import it to a component, I did not use it. Same with the error state.

src/components/Posts/ExpandedPostItem.tsx
NasihNazeem: yeah good shout, i think it does

src/utilities/useAxios.ts
Dennull: I guess you're right because we only use this as a wrapper to send requests right now. If it returned data, loading, and error states that would be a different story right because then we'd need to initialize the hook for each endpoint because we're storing states associated with each endpoint in the hook

Anyway, can we not use `SendRequestProps` and instead pass properties those directly into the function? I don't want to pass a big object into `sendRequest`, I wanna pass just individual parameters. So maybe something like this

```
const sendRequest = useCallback(
    async (endpoint: string, config: AxiosRequestConfig): Promise<any> => {
...
```

src/utilities/useAxios.ts
Dennull: Ok I think you're right because if you remove `useCallback` we run into some infinite render situations from the function being in the `useEffect` dependency array. Ideally we move the `useEffect` into this function and make it fetch on its own. Or alternatively we can add a manual request flag so that we don't need to use `useCallback` on manual requests. But we'll keep it moving for now, since we have future ideas of further improving this hook

Just wanted to point out though that using memoization like `useMemo` and `useCallback` isn't like free optimization. It comes at a cost, so we shouldn't be using it just in case or a precaution. I think the best practice is to use it only when we really need. There should be a reason we use it, and in this case there is, but that's why I asked about if we truly need it or not

src/utilities/useAxios.ts
Dennull: Yeah I'm fine with making this a separate story 👍

src/utilities/useAxios.ts
```diff
@@ -0,0 +1,45 @@
...
+    async ({ endpoint, config }: SendRequestProps): Promise<any> => {
+      try {
+        const token = await getAccessTokenSilently();
+
+        if (!import.meta.env.VITE_BASE_URL) {
+          throw new Error("Base URL is not defined");
+        }
+
+        const response = await axios.request({
+          method: config.method,
```
Dennull: Should we spread the config here so that we automatically get all of the options that come with axios? We currently don't use any of the other ones, so if you wanna leave it for another time that's fine with me too

Dennull (changes requested):

src/utilities/useAxios.ts
NasihNazeem: I removed the `SendRequestProps` type and passed the properties directly to sendRequest. This is a short-term solution as discussed offline, as the long-term fix would be to pass the parameters from the hook rather than the function. 

This means we will occasionally have multiple instances of the `useAxios` hook in components, where each instance has its own `sendRequest` function, along with its respective `data, loading, error` states. We will open up a new ticket to get the long-term fix sorted.

Dennull (approved):

PR #130, opened 2024-10-04
T129: Fix edit post modal crash

### LINKED ISSUE #129: App crashes when on expanded post due to edit post modal
When navigating to an expanded post, the app crashes. Opening the console shows this error:

Image

This is likely because of a `.find()` operation we're doing in the expanded post modal to find the existing post returning `undefined`

Dennull:
## Related Issues

Resolves #129 

## Summary

**Context**

When navigating to an expanded post, the app crashes. This was because we were trying to find the post to pass into our post edit modal using a `.find()` in our `posts` state in the post store. On the expanded post page, this state gets overridden to hold the expanded post replies rather than the previous post list that contained the post the user clicked on; hence, the `find()` operation would return `undefined` because that post id was no longer present in the array

**Solution**

To address this, we shouldn't be looking for the post in the store; rather, we should just pass it down as a prop to the post menu and edit post modal directly

## Changes Made

Passed the post object down as a prop to our `PostMenu` and `EditPostModal` respectively

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes

### DIFF
--- a/src/components/Posts/ExpandedPostItem.tsx
+++ b/src/components/Posts/ExpandedPostItem.tsx
@@ -34,7 +34,6 @@ import UserAvatar from "../Common/UserAvatar";
 import { useTheme } from "@mui/material/styles";
 import PostMenu from "./PostMenu";
 import TooltipTimestamp from "../Common/TooltipTimestamp";
-import formatTimestamp from "../../utilities/formatTimestamp";
 import { enqueueToast } from "../../state/slices/toastSlice";
 
 const styles = {
@@ -124,13 +123,7 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
       </Box>
       <CardHeader
         avatar={<UserAvatar username={post.username} />}
-        action={
-          <PostMenu
-            authorId={post.userId}
-            postId={post.postId}
-            isExpandedPost
-          />
-        }
+        action={<PostMenu authorId={post.userId} post={post} isExpandedPost />}
         title={
           <Link
             color={theme.typography.subtitle1.color}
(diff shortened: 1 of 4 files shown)


src/components/Posts/ExpandedPostItem.tsx
```diff
@@ -124,13 +123,7 @@ const ExpandedPostItem = ({ post }: ExpandedPostItemProps) => {
...
       <CardHeader
         avatar={<UserAvatar username={post.username} />}
-        action={
-          <PostMenu
-            authorId={post.userId}
-            postId={post.postId}
-            isExpandedPost
-          />
-        }
+        action={<PostMenu authorId={post.userId} isExpandedPost post={post} />}
```
NasihNazeem: To avoid redundancy and since we are already passing the entire post object, let's remove the `authorId` and use `post.userId` within the component instead

src/components/Posts/PostItem.tsx
```diff
@@ -85,7 +85,7 @@ const PostItem = ({ post }: PostProps) => {
     <Card sx={styles.card}>
       <CardHeader
         avatar={<UserAvatar username={post.username} />}
-        action={<PostMenu authorId={post.userId} postId={post.postId} />}
+        action={<PostMenu authorId={post.userId} post={post} />}
```
NasihNazeem: Similar comment as my initial comment about the redundancy here. Since we are already passing `post.userId` as part of the `post` object we don't need the `authorId` to be a separate prop here

NasihNazeem (approved):
One small change requested, I'm gonna approve though because it should be good aside from that

src/components/Posts/ExpandedPostItem.tsx
Dennull: Great idea, I'll do this!

PR #133, opened 2024-10-14
T132: Update libraries

### LINKED ISSUE #132: Update front end dependencies

Dennull:
## Related Issues

Resolves #132

## Summary

This PR updates our front end libraries to the latest versions. A few libraries that we don't use were also removed

## Changes Made

All of our libraries were updated to the latest version using `npm update`. This command does **not** update packages with new major semantic versions (ex. going from 2.x.x to 3.x.x), since these changes are breaking, so those had to be done manually

For libraries with major semantic version updates, the docs were referenced and breaking changes were addressed accordingly. Some libraries that we don't use were removed as well. The libraries with major version changes and the libraries that were removed are listed below

### Libraries with Major Updates

auth0-react

- v2 -> v3: https://github.com/auth0/auth0-react/blob/main/MIGRATION_GUIDE.md
- A `authorizationParams` prop was introduced to the provider. Basically, we need to specify some of the existing `Auth0Provider` props such as `audience` and `redirectUri` in a new object prop called `authorizationParams`
- Some of the properties in `authorizationParams` are now snake case (ex. `redirectUri` is now `redirect_uri`)

@mui/material, @mui/icons-material

- v5 -> v6: https://mui.com/material-ui/migration/upgrade-to-v6/, https://mui.com/material-ui/migration/migrating-from-deprecated-apis/
- `Autocomplete` now has some new props we should check out, we may be able to use them
- The `color` prop on `Typography` is not a system prop anymore. According to [the docs](https://mui.com/material-ui/migration/upgrade-to-v6/#typography), some system colors are still accessible using this prop, so it was a bit confusing on what this means.. I think it means that you can't access all theme colors through this prop anymore, only the ones they provide? Either way, I don't think anything broke for us
- A lot of things were [deprecated](https://mui.com/material-ui/migration/migrating-from-deprecated-apis/#paragraph) in this release too. Deprecated APIs won't break anything for now, but support for them are often removed in the next major release version, so we should slowly migrate away from them
  - The big thing that affects us is that they are changing how [inner element overrides](https://mui.com/material-ui/migration/migrating-from-deprecated-apis/#inner-element-overrides) work. TLDR, we have to swap from using the `InputProps` prop to using the new `slotProps` prop instead. I have changed some of our `InputProps` to use their new API.
    - A few other props were affected, such as the old `components` prop and `componentsProps` prop, as shown in their docs. I don't think we use many of these other than `InputProps`, but this is just something to look out for
  - Their [system props](https://mui.com/material-ui/migration/migrating-from-deprecated-apis/#system-props) are also deprecated in favor of the `sx` prop. I don't think we use many of these, but this is also something to look out for

@types/node

- I couldn't really find docs for this but I think nothing is broken lol

typescript

- v4 -> v5: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/

vite

- v3 -> v4: https://v4.vitejs.dev/guide/migration.html
- v4 -> v5: https://vite.dev/guide/migration

mui-x-date-picker

- v5 -> v6: https://mui.com/x/migration/migration-pickers-v5/
- v6 -> v7: https://mui.com/x/migration/migration-pickers-v6/
- The `value` prop has to be of the same type as your date adapter now. For us, we need to wrap our date objects in the `datejs()` parser. Also had to install `datejs` library
- The `renderInput` prop is deprecated, we should use the `slotProps` prop now

reduxjs/toolkit, react-redux

- redux toolkit v1 -> v2
- react-redux v8 -> v9
- These were done at the same time: https://redux-toolkit.js.org/usage/migrating-rtk-2

### Libraries Removed

- web-vitals
- testing-library/jest-dom
- testing-library/react
- testing-library/user-event
- @types/jest
- redux
  - `redux-core` is already bundled in with `redux-toolkit`, so we don't need this
- redux-thunk
  - `redux-thunk` is already bundled in with `redux-toolkit`, so we don't need this

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


NasihNazeem (approved):
I couldn't find anything app-breaking, and I took a look at all the changes and agree with the implementation and the provided documentation.

PR #134, opened 2024-10-18
T131: User Profile Bio Section in Direct Messages

### LINKED ISSUE #131: Direct message bio front end
It'd be nice to have a sort of section in the top portion of a direct message that displays the other person's bio when starting a brand new conversation with someone. This section should display basic information from the user such as their display name, profile picture, username, bio, follower count, and birthday. The section should also be a link that the user can click to navigate to that user's profile

NasihNazeem:
## Related Issues

Closes #131.

## Summary

This PR adds a clickable user bio section within the direct messages page.

## Changes Made

There are modifications to `formatTimestamp()`, `ConversationList.tsx`, `SelectedUser` type and the addition of the Bio Details component. This includes functionalities such as:

- Profile Avatar
- Display Name and Username
- User Bio
- Follower Count
- Navigation (navigates to the user's profile page)
- Added loader for the component
- Modified the SelectedUser type to have optional attributes (bio, joinedDate and followerCount)
- Removed the state setter for the selected conversation inside of `ConversationList.tsx`.
- Addition of an empty string edge case for the `formatTimestamp()` method.

This feature helps fledge out the messages pages, by allowing users to quickly view and access other users' profile details within the chat.

## Screenshots

{1A97BAA0-B745-4C00-BB70-1FC6300AEA66}

## Testing Instructions

No additional notes


src/components/Messages/ConversationList.tsx
```diff
@@ -29,18 +29,26 @@ const ConversationList = () => {
     (state) => state.messages
   );
   const [messageModal, showMessageModal] = useState(false);
+  const [loading, setLoading] = useState(false);
```
Dennull: This loading state doesn't get used anywhere right now. Were you trying to get rid of a flicker or something?

Side note, but I think it'd be nice to have our requests handle the loading state automatically similar to how react query does it rather than defining these loading states everywhere. That can probably be part of our bigger story to add loaders across the app

src/pages/DirectMessage.tsx
```diff
@@ -201,9 +251,40 @@ const DirectMessage = () => {
...
+                      </Typography>
+                      <Typography variant="body2">
+                        {`@${selectedConversation.username}`}
+                      </Typography>
+                    </Box>
+                    <Typography>{selectedConversation.bio}</Typography>
+                    <Typography variant="body2">
+                      {`Joined ${formatTimestamp(
+                        selectedConversation.joinedDate ?? ""
+                      )} •
```
Dennull: Can we make this conditional instead? So if there is no `joinedDate` then we just don't render it

src/utilities/formatTimestamp.ts
```diff
@@ -1,4 +1,6 @@
 const formatTimestamp = (entry: string, isEdited?: boolean) => {
+  if (!entry) return "N/A";
```
Dennull: To go along with the comment I made in `DirectMessage.tsx`, I feel like we should conditionally render timestamps instead of handling empty values in here. The `entry` parameter can't be undefined, and I think it's safe to assume we won't be passing empty strings in

I'm thinking if in the future more cases pop up where we have to handle empty timestamps, then maybe we can deal with that when it comes. Do you agree?

src/pages/DirectMessage.tsx
```diff
@@ -29,8 +31,36 @@ import UserAvatar from "../components/Common/UserAvatar";
 import EmojiPickerIconButton from "../components/Common/EmojiPickerIconButton";
 import { EmojiClickData } from "emoji-picker-react";
 import TooltipTimestamp from "../components/Common/TooltipTimestamp";
+import formatTimestamp from "../utilities/formatTimestamp";
+import PageLoader from "./PageLoader";
```
Dennull, line 36: This is currently not used. Is there a reason we're using `CircularProgress` over our component? Is it the styling?

src/pages/DirectMessage.tsx
```diff
@@ -201,9 +251,40 @@ const DirectMessage = () => {
...
             <List component="div" ref={messageRef} sx={styles.messageList}>
+              <Box
+                sx={styles.bioContainer}
+                onClick={() => navigate(`/${selectedConversation.username}`)}
+              >
+                {loading ? (
+                  <CircularProgress />
+                ) : (
+                  <>
+                    <Box sx={styles.bioDetails}>
```
Dennull: Do we need this div?

src/pages/DirectMessage.tsx
```diff
@@ -201,9 +251,40 @@ const DirectMessage = () => {
...
+                      <Avatar sx={styles.avatar} />
+                    </Box>
+                    <Box sx={styles.nameContainer}>
+                      <Typography variant="h6" sx={styles.displayName}>
```
Dennull: Instead of styling this to look like a link do you think it'd be better if we use a `Link` component? Similar to what we do elsewhere in the app, for example in the post list

src/pages/DirectMessage.tsx
```diff
@@ -63,13 +97,20 @@ const styles = {
...
-  messageList: { flex: 1, overflowY: "scroll" },
+  messageList: { flex: 1, overflowY: "scroll", paddingTop: 0 },
   messageText: {
     padding: 1,
     borderRadius: 10,
     backgroundColor: "primary.light",
   },
   middleContent: { flex: "0 0 350px", height: "100vh", minWidth: 0 },
+  nameContainer: {
+    width: "100%",
```
Dennull: I don't think we need this

src/pages/DirectMessage.tsx
```diff
@@ -201,9 +251,40 @@ const DirectMessage = () => {
...
+                    </Box>
+                    <Box sx={styles.nameContainer}>
+                      <Typography variant="h6" sx={styles.displayName}>
+                        {selectedConversation.displayName}
+                      </Typography>
+                      <Typography variant="body2">
+                        {`@${selectedConversation.username}`}
+                      </Typography>
+                    </Box>
+                    <Typography>{selectedConversation.bio}</Typography>
```
Dennull: Can we make this conditional since not everyone has a bio? I'm not sure if it affects any styling but just in case

src/pages/DirectMessage.tsx
```diff
@@ -201,9 +251,40 @@ const DirectMessage = () => {
...
+
           <Box sx={styles.chatContainer}>
             <List component="div" ref={messageRef} sx={styles.messageList}>
+              <Box
+                sx={styles.bioContainer}
+                onClick={() => navigate(`/${selectedConversation.username}`)}
+              >
+                {loading ? (
+                  <CircularProgress />
+                ) : (
```
Dennull: I'm not really against leaving it like this, but I think this conditional is sort of ugly here. Do you think it'd look nicer if we do 2 conditional renders rather than the ternary?

```
{loading && <CircularProgress />}
{!loading && (
  ... // Bio stuff
)}
```

src/pages/DirectMessage.tsx
Dennull, line 1: Sort of some nitpicks about the styling, I feel like there isn't quite enough space between the display name and username. I also think there's a bit too much space between the profile stuff and the bio. Maybe try 12 pixels or 8 pixels instead?

We can also hop on a call to fiddle with the styling together if you want

image

Dennull (changes requested):

src/components/Messages/ConversationList.tsx
NasihNazeem: >This loading state doesn't get used anywhere right now. Were you trying to get rid of a flicker or something?

Yes I was trying to get rid of a flicker, but I don't seem to be using it so I've removed the state.

Yeah I think the loader states are kind of an inconvenience at the moment. We might need to use something like react query or make something similar to it

src/pages/DirectMessage.tsx
NasihNazeem: Since I need to render both the joinedDate and followerCount what about something like this?

```
{selectedConversation.joinedDate &&
  selectedConversation.followerCount &&
  `Joined ${formatTimestamp(
    selectedConversation.joinedDate
  )} • ${selectedConversation.followerCount} Followers`
}
```

src/utilities/formatTimestamp.ts
NasihNazeem: >I'm thinking if in the future more cases pop up where we have to handle empty timestamps, then maybe we can deal with that when it comes. Do you agree?

Yeah I'm cool with this. This solution felt kinda weird to implement tbh because we didn't do it with alot of the other functions, but I agree if there are too many implementations of the conditional render then we should look to change the solution

src/pages/DirectMessage.tsx
NasihNazeem: It was the styling initially, but because I wrapped the implementation in a box with style `bioContainer`, the styling shouldn't affect it. I will revert to using `PageLoader` as implemented throughout the app

src/pages/DirectMessage.tsx
NasihNazeem: I like this suggestion, it helps improve the readability in the page.

src/pages/DirectMessage.tsx
NasihNazeem: Done

src/pages/DirectMessage.tsx
NasihNazeem: Removed it, it was unneccessary

src/pages/DirectMessage.tsx
Dennull: As discussed offline, it doesn't really make sense to have an undefined joinedDate, so we need to always render that. I see you've made the changes tho, nice 👍

I made a small syntax adjustment to help readability

src/pages/DirectMessage.tsx
Dennull: Ok nice, maybe we should rename `PageLoader` to just `Loader` sometime lol. Probably can do that in the loader ticket

src/pages/DirectMessage.tsx
Dennull: I removed the hover color to make it consistent with the rest of our links

Dennull (approved):

PR #135, opened 2024-11-04
T98: Configure ESLint

### LINKED ISSUE #98: Set up ESLint on frontend
Set up ESlint for our project and look into a plugin that will sort our imports for us. Also configure eslint to catch imports that are out of order so we'll get compiler warnings

Dennull:
## Related Issues

Resolves #98 (finally!)

## Summary

We finally set up ESLint, better late than never lol 😅. You can run the linter using the `npx eslint` command, and automatically fix linting issues using `npx eslint --fix`

## Changes Made

ESLint was configured along with some extra plugins. I ended up going with the newest version of ESLint, v9, because v8 [support ended recently](https://eslint.org/blog/2024/09/eslint-v8-eol-version-support/). More details can be seen on the [ESLint docs](https://eslint.org/docs/latest/). All file changes made in this PR were to fix linting errors, no logic or UI changes were made

Here are some plugins that were installed:
- `eslint-plugin-prettier` was installed to flag formatting during linting. Many formatting issues can be automatically fixed by running `npx eslint --fix`. It will help keep our formatting more tidy and consistent
- `eslint-plugin-import-x` was installed to help sort our imports. Imports that are out of order or not imported properly will be flagged by the linter
- `eslint-plugin-react` was configured to sort our props alphabetically

I also wanted to install `eslint-plugin-react-hooks`, but their ESLint v9 support isn't very good yet.. They technically support ESLint v9 as of this PR as can be seen [here](https://github.com/facebook/react/pull/28773), but they don't have an easy way to use the new [flat config](https://github.com/facebook/react/issues/28313) system yet. There is an open issue about supporting flat config, but it hasn't been closed yet. The docs haven't been updated yet either, so I decided to just leave this one out until they give an update

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


src/components/Common/EmojiPickerIconButton.tsx
```diff
@@ -1,7 +1,7 @@
-import { Box, IconButton } from "@mui/material";
 import EmojiEmotionsOutlinedIcon from "@mui/icons-material/EmojiEmotionsOutlined";
-import React, { useState, useEffect, useRef } from "react";
+import { Box, IconButton } from "@mui/material";
 import EmojiPicker, { EmojiClickData, EmojiStyle } from "emoji-picker-react";
+import React, { useState, useEffect, useRef } from "react";
```
NasihNazeem, line 4: Do we care if the imports within the objects are also sorted alphabetically?

src/components/Posts/PostMenu.tsx
```diff
@@ -50,25 +50,25 @@ const PostMenu = ({ isExpandedPost = false, post }: PostMenuProps) => {
...
+      <IconButton onClick={() => setMenuOpen(true)} ref={menuRef}>
         <MoreVert />
       </IconButton>
       <Menu
         anchorEl={menuRef.current}
-        open={menuOpen}
+        MenuListProps={{ sx: styles.menuList }}
         onClose={() => setMenuOpen(false)}
+        open={menuOpen}
         PaperProps={{
```
NasihNazeem, line 61: unrelated but with the recent upgrade for MUI, I think we should open another ticket to update some of the older implementations

src/pages/Register.tsx
```diff
@@ -58,8 +58,8 @@ const Register = () => {
 
   return (
     <form onSubmit={submitUserInfo} style={styles.container}>
-      <Typography variant="h1" sx={styles.title}>
-        Let's get to know a little more about you
+      <Typography sx={styles.title} variant="h1">
+        {`Let's get to know a little more about you`}
```
NasihNazeem, line 54: is this change something the formatter handles or did you manually edit this?

src/state/slices/messagesSlice.ts
```diff
@@ -24,7 +24,7 @@ const initialState: ConversationDetails = {
   conversations: [],
 };
 
-export const messagesSlice = createSlice({
+const messagesSlice = createSlice({
```
NasihNazeem, line 34: nice, i didn't realize we were unneccesarily exporting the slices

src/state/slices/postsSlice.ts
```diff
@@ -106,22 +106,27 @@ export const postsSlice = createSlice({
       if (action.payload === state.expandedPost.postId) {
         const isLikedByCurrentUser = !state.expandedPost.isLikedByCurrentUser;
         state.expandedPost.isLikedByCurrentUser = isLikedByCurrentUser;
-        isLikedByCurrentUser
-          ? state.expandedPost.numberOfLikes++
-          : state.expandedPost.numberOfLikes--;
+        if (isLikedByCurrentUser) {
```
NasihNazeem, line 109: I've noticed a few turnary statements replaced by if else statements. Is this something the formatter has done or your own edits?

src/index.tsx
```diff
@@ -23,5 +21,5 @@ root.render(
         </Provider>
       </AuthProvider>
     </BrowserRouter>
-  </React.StrictMode>
+  </React.StrictMode>,
```
NasihNazeem, line 24: Is there any way for us to remove the comma that the formatter implements after JSX elements? I tried looking it up and I think we have to make a change to the `"trailingComma"` setting inside of the `.prettierrc` file

NasihNazeem (commented):
Great job putting this together! I left a few comments/questions before merging this through. 

I love the changes though, it looks much cleaner than it was before

src/components/Common/EmojiPickerIconButton.tsx
Dennull: I don't think the plugin supports that, so unfortunately we might have to leave it. Feel free to browse their docs, maybe I missed it https://github.com/un-ts/eslint-plugin-import-x

src/components/Posts/PostMenu.tsx
Dennull: Yea you're right, I made a new ticket

src/pages/Register.tsx
Dennull: It's actually not the formatter, it's `eslint-plugin-react`. It's because of the apostrophe character, [one of the rules](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md) prevents escape characters from appearing in the body of a tag without using special codes

image

src/state/slices/postsSlice.ts
Dennull: This is a [rule](https://typescript-eslint.io/rules/no-unused-expressions/) in `typescript-eslint`. I think the point is that unused expressions like this can get confusing? Sort of a best practice thing I think. I can't find a solid answer but I think people like using ternaries for assignment operations and not side effects like increment/decrement or function calls

I'm personally ok with keeping the rule as it is, but we can disable the rule or make it allow for ternaries if we want

```
"@typescript-eslint/no-unused-expressions": [
  "error",
  { allowTernary: true },
],
```

src/index.tsx
Dennull: I read up on why this is the default in the first place and it looks like it's for consistency, cleaner code reviews (the diffs apparently look better), and easily moving stuff around. For example, if you want to move items between a list around or parameters for functions around having the trailing comma makes it so that you can cut and paste the new line without having to add a new comma

It turns out this comma isn't on trailing JSX; it's because the JSX is in a function (`root.render()`). We are able to set `trailingComma` to be `es5` in the `.prettierrc` file to make it so that we have trailing commas on arrays but not function parameters. I don't really mind having the commas, but if it really bothers us I'm not against setting it to `es5`. But now knowing all that stuff do you think you want to change it?

eslint.config.mjs
```diff
@@ -0,0 +1,42 @@
...
+  pluginReact.configs.flat["jsx-runtime"],
+  eslintPluginPrettierRecommended,
+  eslintPluginImportX.flatConfigs.recommended,
+  eslintPluginImportX.flatConfigs.typescript,
+  {
+    languageOptions: {
+      parser: tsParser,
+      ecmaVersion: "latest",
+      sourceType: "module",
+    },
```
NasihNazeem, line 26: When I try running `npx eslint` or `--fix` I get this warning log that says React version is not specified in eslint-plugin-react settings. See https://github.com/jsx-eslint/eslint-plugin-react#configuration

{7A26F27D-B4AD-4C99-AFB7-75E0054A3FEA}


I looked into it myself and I think we have to add the following object

```suggestion
    },
   settings: {
      react: {
        version: "detect",
      },
    },
    rules: { ...
```

eslint.config.mjs
NasihNazeem, line 1: When I tried running the command with the provided solution (settings object), I noticed that the command took quite some time, so I chat gpt'd reasons as to why it may be, and I guess I concluded with the idea that when I run the command in the root directory, it goes through all folders and files. I'm not sure if that's the reason why it takes so long, but I tested it by forcing the linting to only work on the src folder and it wound up being alot quicker

`npx eslint --fix src/` 

Do you find a similar issue on your end or is this isolated to my computer?

NasihNazeem (changes requested):
Good work so far! I was testing out the linting command and there was a minor warning I found, so I suggested a change for that. There was also a random "bug" I found that I explained in another comment where I reference the suggested solution.

eslint.config.mjs
Dennull: Oh I guess I just ignored this warning lol. Nice 👍

eslint.config.mjs
Dennull: I do find it kind of slow (like 10 seconds slow), but when I run `npx eslint --fix src/` the time is about the same for me. I suspected that maybe it was linting node modules but by default eslint ignores node modules. I also ran `npx eslint --debug` and confirmed so that node modules wasn't being linted

I guess `eslint-plugin-prettier` adds some overhead, but when I tried removing the it the time was still was >5 seconds, so I think it's also `eslint-plugin-import-x`. Some files just take a while to lint with the plugins. You can try running `npx eslint --debug` yourself to see how long stuff takes to run, I'm open to any suggestions. Given that we're probably only running this once in a while I'm inclined to move forward and work on continue working on improving the performance later though, let me know if you agree

Dennull: Okay, as discussed offline, it turns out the issue was that the linter was looking through your build folder. I managed to find out why my config to ignore the build directory wasn't working and fixed it. Give it a shot and see if that stops the linter from going forever 👍

NasihNazeem (approved):

NasihNazeem (approved):
I reviewed all changed files and nothing popped out at me in terms of a potential bug. I did notice an outdated PaperProps in `PostMenu.tsx` that should be changed to a `slotProps` but I think I'll keep that in mind for the next PR. It's not breaking anything.

I also tested out my eslint and it wasn't working initially, I ran a --debug and tried again and it worked as expected! I'm not sure what the --debug did though because it ran through a ton of files with logging going past the accessible amount of lines in my terminal.

We can finally merge this, let's go!!!

Issue #140, opened 2025-01-01
Consolidate typing


Dennull:
Our typing, specifically for our user and message types, are a bit of a mess right now. A few things that can be cleaned up include:

- `SelectedUser` and `Conversation` have a lot of overlap, we can probably make one inherit from the other rather than duplicating typing
- `SelectedUser` has some properties that are used only for messaging purposes. For example, the `bio` property is only used for the message bio component. In other places where `SelectedUser` is used, such as `SearchBarDropdown.tsx`, these properties aren't needed
-  The `NetworkUsers` type is very similar to `SelectedUser`. Maybe we can consolidate both of these into one type or make the difference more clear
- Consider creating separate type files to put our types in, or at least move all of our types into the relevant slices

Overall, while we don't have solid answers to these problems, it's worth exploring how we can set our typing straight in this ticket. We should aim to make our typing more clear so that it's easier to know expect what each component needs and what we expect the back end to return

Issue #138, opened 2024-12-13
Live messaging


Dennull:
This ticket is to implement real-time messaging. While we currently have a REST API to get messages and store them in the database, this functionality currently isn't live; users must refresh the page to get the latest messages. We will likely have to use WebSocket to achieve the desired functionality. Additionally, we'd also like to implement in-app notifications so that users are aware when they've received a new message, provided they're logged in. MVP requirements:

- Users should be able to send and receive messages in real-time
- A badge should appear on the messages button in the navbar showing how many new messages a user has
  - This should be dismissed when users view the message
  - The number in the badge should display the number of unread conversations in their chats
- Unread conversations should be highlighted (the text bolded) to show that the user has received unread messages from that conversation
  - Also add a badge at the end of the conversation item such as a blue dot to add some more indication that the conversation has new messages

There are also a few more optional nice to haves to explore as part of this ticket. These are more investigations, and should be done if they're easy. If not, we can make a separate ticket to add them:

- Add an animation or icon if the other person is currently typing in the chat
- Add a "delivered" and "seen at" flag

PR #141, opened 2025-01-04
T139: Chat refactor

### LINKED ISSUE #139: Chat component refactor in direct message page
The direct message page can be improved on by refactoring out the chat section (the area where the list of messages are rendered) into new components. This will likely have to come with moving the messages state into the store

Dennull:
## Related Issues

Resolves #139 

## Summary

`DirectMessage.tsx` has long needed a refactor to make it simpler, since right now we jammed a bunch of stuff in there. Because our upcoming live messaging feature will likely change the way our messaging words significantly, these refactors were done as a prerequisite for #138

## Changes Made

The main changes in this PR are that the stuff in `DirectMessage.tsx` was broken up into a few smaller components:

- `ChatContainer.tsx` holds the general chat window, including the header, the chat list, and the input
- `ChatList.tsx` holds the list of messages along with the chat bio
- The chat bio was made into it's own component, `ChatBio.tsx`
- The chat input got put into `ChatInput.tsx` to prevent the entire chat list from rerendering whenever the user typed into the input because this was causing some lag

A few other auxiliary changes were made while I was at it:

- A bug that caused the selected item on the navbar to flicker when the user swapped between conversations was fixed
  - The bug was because we determined the selected button based on an exact match of the routes. The route would change every time the user clicked on a conversation because of the user ids in the URL
  - To fix the bug, the selected button was changed to be based on a match of the base path of the route instead (ie. so anything with `/messages/` will highlight the messages nav item)
- The `updateConversation` and `appendConversation` reducers were combined and simplified into a new one, `addMessage`
- A few types were changed and simplified, but this is to be further expanded on in #140
- Some API endpoint names and returns were changed for better clarity

All of this was bundled with general cleanup of our code in the message components. There was no intention of changing any behavior other than the bug with the navbar flicker, so all of our features other than that bug should remain the same

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


src/components/Messages/ChatContainer.tsx
```diff
@@ -0,0 +1,76 @@
...
+    display: "flex",
+    justifyContent: "space-between",
+    paddingLeft: 2,
+    paddingRight: 2,
+  },
+  headerContent: { alignItems: "center", display: "flex", gap: 2, padding: 1 },
+};
+
+const ChatContainer = () => {
+  const { messages, selectedConversation } = useAppSelector(
```
NasihNazeem, line 26: maybe we can split these two into two `useAppSelector` calls?

src/components/Messages/ChatContainer.tsx
```diff
@@ -0,0 +1,76 @@
...
+import UserAvatar from "../Common/UserAvatar";
+import ChatInput from "./ChatInput";
+import ChatList from "./ChatList";
+
+const styles = {
+  headerContainer: {
+    alignItems: "center",
+    display: "flex",
+    justifyContent: "space-between",
+    paddingLeft: 2,
```
NasihNazeem: `paddingX` instead of left and right?

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
```diff
@@ -11,7 +11,7 @@ type MessagesListProps = {
 
 const MessagesList = ({ onClose }: MessagesListProps) => {
   const userId = useAppSelector((state) => state.user.userId);
-  const [conversationList, setConversationList] = useState<SelectedUser[]>([]);
+  const [conversationList, setConversationList] = useState<Conversation[]>([]);
```
NasihNazeem, line 14: Do we need type Conversation here? Is it because of the query return? SelectedUser might be enough i think

src/components/Messages/ChatContainer.tsx
```diff
@@ -0,0 +1,76 @@
...
+              {selectedConversation.displayName}
+            </Typography>
+            <Typography variant="subtitle2">{`@${selectedConversation.username}`}</Typography>
+          </Box>
+        </Box>
+        <IconButton>
+          <InfoOutlinedIcon />
+        </IconButton>
+      </Box>
+      <ChatList messages={messages} userDetail={selectedConversation} />
```
NasihNazeem, line 68: What do you think of accessing the store for this within the component itself rather than passing it as a prop

NasihNazeem (changes requested):
A few nitpicks, thank you for cleaning this page up completely

src/components/Messages/ChatContainer.tsx
Dennull: I think I want to leave this for the redux ticket to address all of these things at once

src/components/Messages/CreateMessageModal/MessagesModalList.tsx
Dennull: I see, I think I saw this comment on one of the other PRs so I'll address it there since this type changed

src/components/Messages/ChatContainer.tsx
Dennull: I thought about that, but I think I went with this because we can potentially reuse this component for a chat drawer. I think if that happens we'd need separate states because you can have two different DMs opened between the messages page and the chat drawer

There could be an argument to be made that this is unnecessary future proofing though, let me know what you think

Dennull: As discussed offline, we'll keep it as it is for if we want to reuse this component in the chat drawer/inbox

NasihNazeem (approved):

PR #142, opened 2025-01-04
T140: Consolidate typing

### LINKED ISSUE #140: Consolidate typing

Dennull:
## Related Issues

Resolves #140 

## Summary

This PR improves the clarity and organization of our TypeScript type definitions by moving types to a dedicated `types/` folder and refining their usage

Types that were previously shared across unrelated components have been reviewed and redefined to better fit their respective contexts. For example, the `SelectedUser` type included properties specific to messaging before, and has been refactored to separate messaging-related properties from those used in non-messaging contexts (such as in the search bar)

Another issue was that a few of our types had significant overlap, which made things seem redundant sometimes. These overlapping types have been consolidated or redefined to minimize redundancy. TypeScript utility types, such as intersection types (`&` operator), have been used to better capture relationships between types, making it clearer when and how to use specific type definitions

## Changes Made

The `SelectedUser` type was renamed to `BaseUser`. This type ended up being used more broadly for situations where we had to display a user's avatar along with their usename and display name. As this happens quite often in the app, and these properties are generally needed whenever we display any user, `BaseUser` feels like a more appropriate name

It generally feels like other user related types inherit from `BaseUser`, so inheritance pattern was followed for the rest of the changes:

- The `NetworkUsers` type in `FollowListModal.tsx` was changed to `FollowableUser`, which inherits from `BaseUser`
- `ProfileContent` extends from `BaseUser` and adds more types. This sort of feels like a "full user", since a user's profile content has everything about them that would be displayed to the public, so we could potentially rename this
- The `Conversation` type extends from `BaseUser` as well now
- Our slice state types were changed to follow a consistent naming pattern

Finally, shared types have been moved into their own folder and files to make them easier to find

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes

### DIFF
--- a/src/state/slices/userSlice.ts
+++ b/src/state/slices/userSlice.ts
@@ -1,5 +1,16 @@
 import { PayloadAction, createSlice } from "@reduxjs/toolkit";
 
+export type AvatarUser = {
+  displayName: string;
+  imageUrl?: string;
+  userId: number;
+  username: string;
+};
+
+export type FollowableUser = AvatarUser & {
+  isFollowing: boolean;
+};
+
 type User = {
   isLoading: boolean;
   userId?: number;
--- a/src/state/slices/userSlice.ts
+++ b/src/state/slices/userSlice.ts
@@ -11,14 +11,14 @@ export type FollowableUser = AvatarUser & {
   isFollowing: boolean;
 };
 
-type User = {
+export type CurrentUser = {
   isLoading: boolean;
   userId?: number;
   username?: string;
   displayName?: string;
 };
 
-const initialState: User = {
+const initialState: CurrentUser = {
   isLoading: true,
 };
 
@@ -29,7 +29,7 @@ const userSlice = createSlice({
     setDisplayName: (state, action: PayloadAction<string>) => {
       state.displayName = action.payload;
     },
-    setUser: (_, action: PayloadAction<User>) => {
+    setUser: (_, action: PayloadAction<CurrentUser>) => {
       return { ...action.payload };
     },
   },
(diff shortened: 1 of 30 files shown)


src/state/slices/userSlice.ts
```diff
@@ -1,13 +1,13 @@
 import { PayloadAction, createSlice } from "@reduxjs/toolkit";
 
-type User = {
+type CurrentUser = {
```
NasihNazeem, line 3: Should we move these types into the types folder as well?

NasihNazeem (changes requested):

NasihNazeem (changes requested):
not many changes requested, great work

src/state/slices/userSlice.ts
Dennull: My idea was that since this type is only being used in the slice we can keep it in there. I was thinking about put only exported/shared types in the types folder, let me know if you agree

NasihNazeem (approved):

PR #144, opened 2025-01-08
T143: Search for all users in new message modal

### LINKED ISSUE #143: Expand new message modal to search for all users
Our new search bar allows for users to search for any user in the app. With this new functionality, we should be able to expand our new message modal's functionality to search for all users rather than followers only

Dennull:
## Related Issues

Closes #143 

## Summary

Before, when creating a new message using the new message modal, the search bar was limited to search for only users that the current user is following. This was originally done to keep things simple. However, with the addition of the `SearchBarDropdown` component, there is now both front end and back end functionality to search for all users across the app. With this being the case, we should expand the new message modal functionality to also search for all users

## Changes Made

- The `SearchBarDropdown` component was renamed to `UserSearchBar`. It was then refactored a bit to be more general so that other components can use it
- The new message modal was refactored to make things simpler, and the existing `MessagesSearchBar` was replaced with `UserSearchBar`

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


src/components/Messages/NewMessageModal/NewMessageModalList.tsx
```diff
@@ -0,0 +1,67 @@
...
+  },
+};
+
+type NewMessageModalListProps = {
+  onSelect: (userId: number) => void;
+};
+
+const NewMessageModalList = ({ onSelect }: NewMessageModalListProps) => {
+  const userId = useAppSelector((state) => state.user.userId);
+  const [conversationList, setConversationList] = useState<Conversation[]>([]);
```
NasihNazeem: I feel like `SelectedUser` type makes more sense here since we dont use timestamp or textContent

src/components/Common/UserSearchBar.tsx
```diff
@@ -108,14 +114,20 @@ const SearchBarDropDown = ({ placeholder }: SearchBarDropDownProps) => {
...
         loading={loading}
         noOptionsText={
-          !keywords ? "Start typing to search..." : "No user found"
+          !keywords ? "Start typing to search..." : "No people found"
         }
-        onBlur={() => setFocusSearchBar(false)}
-        onChange={(_, value) => value && handleSelect(value.username)}
-        onFocus={() => setFocusSearchBar(true)}
+        onBlur={() => {
+          onBlur?.();
```
NasihNazeem, line 123: If we are calling `setFocusSearchBar` within here, i dont think we need this call. Let me know what you think or if i am missing something

NasihNazeem (changes requested):

src/components/Messages/NewMessageModal/NewMessageModalList.tsx
NasihNazeem: Additionally it would be called something other than `conversationList if we change the type

NasihNazeem (changes requested):
Good changes here, I think I am misunderstanding a couple changes but i left a comment anyways cuz I think we can improve our implementation for search bar focus.

Great work Dennis!

src/components/Messages/NewMessageModal/NewMessageModalList.tsx
Dennull: Oh yea I see what you mean, I think the reason is because of what the endpoint returns. We use the same endpoint to get the conversation list (in `ConversationList.tsx`), so I reused the type

Do you think it's worth making a new endpoint to return less data? Or I could also stick to using the endpoint and map only the properties we need

src/components/Common/UserSearchBar.tsx
Dennull: This is for the parent component, this `focusSearchBar` state is for the dropdown to open and styling, the other one in `NewMessageModal.tsx` is for whether or not to display the list of current messages. I can maybe look at changing the name of one of them to make it more clear tho

src/components/Messages/NewMessageModal/NewMessageModalList.tsx
NasihNazeem: Discussed offline, we will turn the conversationList into a prop of type `BaseUser` and remove the endpoint as it is no longer required.

src/components/Common/UserSearchBar.tsx
NasihNazeem: Bit of a misunderstanding on my part. We need both the onBlur and the blur state for the search bar within the modal. I was suggesting that we could merge the two, however, the states are used for two separate components. We discussed offline that we would differentiate the names of the state to make it more clear.

NasihNazeem (approved):

PR #148, opened 2025-03-17
T147: Various Redux optimizations

### LINKED ISSUE #147: Redux optimizations
A few Redux optimizations could be made to improve the performance of the app. In general we shouldn't need to optimize things for no reason, but it might be good to take a look at the Redux best practices and make sure we're following them. A few examples of things that can be done:

- Extract only properties each component needs from the state to prevent rerenders
- Derive more values from the state itself using the selectors (ex. grabbing the length of an array from within the selector rather than grabbing the array from the selector then computing the length)
- Memoizing and/or normalizing lists

Dennull:
## Related Issues

Resolves #147 

## Summary

Some optimizations to how we use Redux. Nothing too crazy was done, I just took a look at some of the Redux best practices and implemented them. Mostly tackled the simpler things to get some potential easy performance wins and improve the codebase a bit. Some of the main sources I used were the [Redux docs](https://redux.js.org/usage/deriving-data-selectors), this [stack overflow post](https://stackoverflow.com/questions/74491856/how-should-i-use-selectors-in-redux-toolkit), and this [video](https://www.youtube.com/watch?v=GdOgYQzGexY)

## Changes Made

- Created centralized selectors by making selector functions in our slices for common use cases. This reduces code duplication and makes it easier to modify things down the line if the shape of our state object changes
- Selected only relevant data that each component needs to prevent unnecessary rerenders
  - Previously, we were grabbing entire objects from our state and destructuring them, which isn't the best thing to do, as if unrelated parts of the state change, the component rerenders when it doesn't need to
- Started using `createSelector` to created memoized selector functions whenever we need to derive state

Further performance optimizations could definitely be made. For example, we could use `React.memo` on our list components to prevent each list item from rerendering unnecessarily if its props don't change. We could also explore if normalizing our state would help us get any performance gains

However, I opted to leave these out for now because I didn't want to do extra work if we don't need it. I think that as a general rule of thumb, we should only spend significant time on optimizations if the app performs noticeably poorly. The performance of the app seems acceptable for now, so I think we should explore other optimizations only if things app starts to feel really slow

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


src/components/NavBar/NavBar.tsx
```diff
@@ -64,8 +65,8 @@ const styles = {
 const NavBar = () => {
   const location = useLocation();
   const [openModal, setOpenModal] = useState(false);
-  const selectedConversationUserId = useAppSelector(
-    (state) => state.messages.selectedConversation.userId,
+  const { userId: selectedConversationUserId } = useAppSelector(
```
NasihNazeem: Since we use this a couple of times, do you think it is worth it to create a selector for the userId belonging to the SelectedConversation user?

src/components/Posts/EditPostModal.tsx
```diff
@@ -66,7 +66,7 @@ const EditPostModal = ({
   open,
   post,
 }: EditPostModalProps) => {
-  const user = useAppSelector((state) => state.user);
+  const username = useAppSelector((state) => state.user.username);
```
NasihNazeem, line 69: I like that you decided it wasn't worth creating a selector here, since we only use it once.

src/state/slices/postsSlice.ts
```diff
@@ -122,6 +123,20 @@ const postsSlice = createSlice({
   },
 });
 
+export const selectPosts = (state: RootState) => state.posts.posts;
+
+export const selectExpandedPost = (state: RootState) =>
+  state.posts.expandedPost;
+
+export const selectNonReplyPosts = createSelector([selectPosts], (posts) =>
```
NasihNazeem, line 131: Nice! I like that we have used `createSelector` here because I think it takes care of the memoization of these arrays since we use the filter() function.

This is my understanding of the following section in the [article](https://redux.js.org/usage/deriving-data-selectors#createselector-overview).

NasihNazeem (approved):
This is great work! I can appreciate the improvements you have made here and it looks great overall. I left a few comments but nothing that requires any real "changes". I would suggest waiting on merging this in, and make sure we merge the hotfix for the workflow file first.

Down the line, I can see us needing a selectors folder and moving the selectors there as it can get cluttered storing everything in a slice. I also wonder if we ever need to have a selector that requires access of state from two slices, then we may need to incorporate like a combinedSelectors.ts file or something. Down the line stuff though I think

src/state/slices/postsSlice.ts
Dennull: Yeah, that's right. Since .filter() returns a copy, the original array reference isn't the same as the old one anymore, so when the selector runs again (every time an action is dispatched), the component will rerender

src/components/NavBar/NavBar.tsx
Dennull: Originally I left it like this cause I didn't think we gain much from it since most of the time, whenever anything from selected conversation changes, we have to rerender anything that uses this anyways. I guess technically it's possible to update that person's display name only or something, but we don't have a use case for that right now; we pretty much always update the entire selected conversation object

I went ahead and made a new selector as suggested anyways tho for consistency sake. It doesn't really hurt anyways, I guess we save some code so it's fine 👍

NasihNazeem:
Cool I think this is good to go

PR #150, opened 2025-03-22
T149: Workflow fix

### LINKED ISSUE #149: Hotfix: actions/cache v2 has been deprecated and needs to be upgraded to v4
The GitHub Actions team has recently announced that they will be deprecating previous versions of cache such as `v1` and `v2`, and have also released their most recent version `v4`. All projects using `v1` and `v2` will have their workflows broken, aside from Enterprise Teams, so we need to update our workflow to reflect this change.

NasihNazeem:
## Related Issues

Resolves #149.

## Summary

We recently ran into an issue with the workflow set up where it would fail during set up. After a little bit of digging, I noticed that the GitHub Actions team has recently [announced ](https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down) updates to the cache repository, where they have released their latest version `v4` and subsequently announced the deprecations of previous versions `v1` and `v2`. In our workflow, we were using `v2` to install our dependencies. 

This means we need to update our cache job to use `v4` as it is the most recent cache storage version released. The workflow file will run our default jobs on a push event or when a pull request is opened.

During testing the proposed change, there was also a warning about another deprecation related to Vite. Vite has [announced ](https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated) that they are moving away from using CommonJS for their Node API and is moving to using an ECMAScript build moving forward. Therefore, this has also been updated in this PR just as an upgrade to a workflow file and the Vite config file.

## Changes Made

- Updated the job that installs dependencies to use the latest cache storage version (v4)
- Renamed the Vite config file to use the ECMAScript build moving forward instead of the CJS build.

## Screenshots

No additional screenshots

## Testing Instructions

No additional notes


Dennull (approved):
Thanks for fixing this!

